PR #4, opened 2017-09-07
Fixed wrong branches coverage

### LINKED ISSUE #3: Issues in branches coverage
I’ve added [jest-vue](https://github.com/eddyerburgh/jest-vue) to my project and I realized that the values for the branches coverage are incorrect. I tried the same with your project [vue-test-utils-jest-example-master](https://github.com/eddyerburgh/vue-test-utils-jest-example) and the issues are the same. 
￼
screen shot 2017-09-07 at 12 28 24

You can notice that the template lines are included in the coverage calculation and it also shows 50% of 6 branches coverage when you don’t even have a single branch in the file… Something is wrong here. Can you look at it?

vanessasml:
@eddyerburgh I think I solved the problem. I realized you were updating the variable beforeLines
before updating the output and because of that when the template parts were added It went wrong and included some of the template lines in the branches coverage.

I've updated the new changes in my project and in yours [vue-test-utils-jest-example](https://github.com/eddyerburgh/vue-test-utils-jest-example) and it works in both projects.
I had to update the sourcemap value in tests for ensure that they would pass.

Please check it ;)

Fixes #3


eddyerburgh:
Looks good @vanessasml 🙂

eddyerburgh:
This fix is live in 0.5.2

PR #12, opened 2017-10-11
Use the absolute path to babel preset

egoist:
let babel use the built-in babel preset instead of looking for it in current working directory (may not exist).


eddyerburgh:
Released in 0.8.1

PR #19, opened 2017-10-26
doc: update changelog

ankurk91:
This PR is an attempt to fix change logs.

**And a question -**

Why should we add this line to changelog ?
```html
<a name="0.7.2"></a>
```
While next line already creates an hyperlink
```
## [Heading](http://url)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,11 @@
 <a name="1.0.0"></a>
-
-Renamed to vue-jest.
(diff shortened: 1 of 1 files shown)


eddyerburgh:
I appreciate the PR, but the changelog is generated automatically. Changing this would mean removing the heading on each release. At the moment the release notes are generated and I just copy and paste. Soon I'd like to prepend the release notes so that it's 100% auromated. So in going to close this pr

ankurk91:
It is ok.

Don't forget fix the version number for `0.8.0`, right now it says **v2.0.0**

screen shot 2017-10-26 at 11 11 28 am

PR #35, opened 2017-12-11
Transpile coffeescript written in ES6

wataruoguchi:
`coffeescript` can adopt babel as a transpiler. `transpile` is the way how they set transpiler.

Please see http://coffeescript.org/#transpilation

This allows you to have vue files which adopt coffee/coffeescript (ES6) as `lang`.
I added tests and vue files.

### DIFF
--- a/lib/compilers/coffee-compiler.js
+++ b/lib/compilers/coffee-compiler.js
@@ -1,5 +1,6 @@
 var ensureRequire = require('../ensure-require.js')
 const throwError = require('../throw-error')
+const loadBabelConfig = require('../load-babel-config.js')
 
 module.exports = function (raw, cb, compiler) {
   ensureRequire('coffee', ['coffeescript'])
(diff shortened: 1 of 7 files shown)


lib/compilers/coffee-compiler.js
```diff
@@ -8,7 +8,10 @@ module.exports = function (raw, cb, compiler) {
   try {
     compiled = coffee.compile(raw, {
       bare: true,
-      sourceMap: true
+      sourceMap: true,
+      transpile: {
+        presets: ['env']
```
eddyerburgh: Would we want to use presets that the user has in their babelrc?

wataruoguchi: @eddyerburgh Yes, actually that should be better

wataruoguchi:
@eddyerburgh Thanks for reviewing the previous commit. These changes should allow users to use their .babelrc

wataruoguchi:
@eddyerburgh Here ^ some tests added. Please review it. Pleasure to contribute to this project :)

eddyerburgh:
Thanks @wataruoguchi 🙂

Issue #36, opened 2017-12-13
Code coverage doesn't work on files linked with "src="


Toilal:
I try to add code coverage to my project, but I use `src=` and it doesn't seem to be supported.

Maybe it's related to sourcemaps missing (https://github.com/eddyerburgh/vue-jest/pull/29#issuecomment-347252861).

eddyerburgh:
Yeah, sorry. Source maps are missing at the moment. I'll add them soon

eddyerburgh:
This was fixed in #131

PR #37, opened 2017-12-13
feat(sourceMap): Make source-mapping of multi-files component working…

csailly:
PR for issue #36 Code coverage doesn't work on files linked with "src="

### DIFF
--- a/lib/process.js
+++ b/lib/process.js
@@ -50,7 +50,12 @@ module.exports = function (src, path) {
   const script = result.code
   const inputMap = result.sourceMap
 
-  const map = generateSourceMap(script, '', path, src, inputMap)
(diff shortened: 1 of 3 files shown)


eddyerburgh:
This is awesome, thanks so much 🙂

Toilal:
It still doesn't work with `lang="ts"` (TypeScript)

dinvlad:
Hmm, it works if I set
```ts
<script lang="ts">
  export { default } from './component';
</script>
```
but not if I use
```ts
<script src="./component.ts"></script>
```
Could this be improved to include the second use case?

eddyerburgh:
@dinvlad could you please open an issue. Then I can keep track of the bug and look into fixing it

PR #43, opened 2018-01-15
Fix Babel 7 compat

yyx990803:
This makes `babel-core` a peer dep so that it can use whatever version of `babel-core` the user specifies.

Also removes the default `babel-preset-vue-app` configuration - I don't think it's a good idea to hard code that in here, especially when we are going to ship a new preset with vue-cli 3.0. If the user doesn't have a `.babelrc`, vue-jest should just skip babel.


eddyerburgh:
Thanks @yyx990803 😀.

I agree with skipping if there's no babelrc.

This fixes https://github.com/eddyerburgh/vue-jest/issues/39

PR #49, opened 2018-01-23
Add support for <style> and the Stylus lang

### LINKED ISSUE #38: Support for CSS modules
As the README states, this lib currently doesn't process the `<style>` parts.

That's not a big issue when we unit test components that use the normal styles API, by which I mean they use harcoded class names in the template:
```html
<div class="staticClass" v-bind:class="{ 'dnyamicClass' : someFlag }">
```

However, for CSS modules, this means that `this.$styles` is undefined in unit tests and rendering will fail.

I think we have two options:

### A: Mock `$styles` 

We could simply add this prop to a component if we find a `style` part with the `modules`attribute, and use a Proxy to return the class that was required:
```
// pseudocode
  this.$style = new Proxy({}, {
    get(_, property) {
      return property
    }
  })
```

Proxies are available in node since http://node.green/#ES2015-built-ins-Proxy, so that's quite good. We are dropping support for node 4.* in vue-cli etc. as well, so that should be enough.

The only downside I see it that we can't catch missing classes: If I mistype a class name in my component, I will still get back that mistyped classname from the proxy. If we properly evaluated the styles, we would get `undefined` and could catch this bug better, I think.

### B: Evaluating styles and building the $styles object.

We could add the usual popular preprocessors as peer deps, write a small wrapper for each to compile the styles, and then use something like [string-extract-class-names](https://www.npmjs.com/package/string-extract-class-names) to get all classnames from the resulting CSS.

From that, we could build a "real" `$styles` object and inject it into the component.

@eddyerburgh - what approach would you prefer?

danielbastos11:
Fixes #38


danielbastos11:
Thanks for the feedback!
I pushed the changes as you requested.
Please let me know if there's anything else.

eddyerburgh:
Thanks 🙂

LinusBorg:
Great Work!

danielbastos11:
@LinusBorg Thanks! :)

lib/process.js
```diff
@@ -78,5 +98,20 @@ module.exports = function (src, path) {
     }
   }
 
+  if (Array.isArray(parts.styles) && parts.styles.length > 0) {
+    const styleStr = parts.styles.map(ast => {
```
LinusBorg: we should only convert style partss that acually have a "module" attribute, right?

Otherwise we would create a `$styles` object on a component that onnly has "normal" styles like `<style lang="scss">`.

lib/process.js
```diff
@@ -78,5 +98,20 @@ module.exports = function (src, path) {
...
+      const moduleName = ast.module || '$style'
+      const styleObj = processStyle(ast)
+
+      return '\n  this[\'' + moduleName + '\'] = ' + JSON.stringify(styleObj)
+    })
+
+    output += '\n;(function() {' +
+      '\nvar render = __vue__options__.render' +
+      '\n__vue__options__.render = function(h) {' + styleStr +
```
LinusBorg: Instead of adding it to the render function, I think we should create a beforeCreate() callback that adds the property during intialization of the component.

That's essentially what vue-loader does, it would look something like this:

```javascript
output += '\n;(function() {' +
      '\nvar existing = ___vue__options__.beforeCreate' +
      '\nvar styleFn =  function () { ' + styleStr + ' })' +
      '\n__vue__options__.beforeCreate = existing ? [].concat(existing, styleFn) : [styleFn]' +
      '})()'
```

This would ensure that the style is also available before render, e.g. to be accessed from within some method or a `created()` hook.

danielbastos11:
Well, I think since neither one of those are use cases I'm currently in need of, I just didn't remember them...
I'll try to update the PR later today to account for these modifications.

Thanks for the review 👍

PR #50, opened 2018-01-24
Styles

### LINKED ISSUE #38: Support for CSS modules

danielbastos11:
I had to open a new PR since #49 had already been merged.
Fixes #38


danielbastos11:
My pleasure :}

PR #54, opened 2018-02-06
Fix issue parsing relative imports in stylus files

### LINKED ISSUE #52: Error parsing stylus that contains imports
## Description
If a Vue template contains a stylus that has an import to an external stylus file, the test suite requiring this file fails with the message:

```shell
failed to locate @import file path/to/stylus/file/in/vuetemplate
```

vue-jest-stylus-issue

## To reproduce the issue
`git clone https://github.com/mdvorscak/vue-test-utils-jest-example.git && npm i && npm run test`

mdvorscak:
This PR fixes #52

### DIFF
--- a/lib/process.js
+++ b/lib/process.js
@@ -28,13 +28,14 @@ function processScript (scriptPart) {
   return compileBabel(scriptPart.content)
 }
 
-function processStyle (stylePart) {
+function processStyle (stylePart, path) {
   if (!stylePart) return {}
 
   let cssCode = stylePart.content
   if (/^styl|stylus$/.test(stylePart.lang)) {
+    const vuePath = join(path, '..')
     const stylus = require('stylus')
-    cssCode = stylus.render(stylePart.content)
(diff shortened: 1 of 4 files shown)


lib/process.js
```diff
@@ -28,13 +28,14 @@ function processScript (scriptPart) {
   return compileBabel(scriptPart.content)
 }
 
-function processStyle (stylePart) {
+function processStyle (stylePart, path) {
   if (!stylePart) return {}
 
   let cssCode = stylePart.content
   if (/^styl|stylus$/.test(stylePart.lang)) {
+    const vuePath = join(path, '..')
```
eddyerburgh: Can you use process.cwd for the path? Like in vueify—https://github.com/vuejs/vueify/blob/master/lib/compilers/stylus.js#L14

mdvorscak:
Thanks for taking a look, I also changed the directory resolution to be clearer. And as a result renamed the path variables to filePath to avoid confusion.

eddyerburgh:
Thanks 😀

PR #58, opened 2018-02-20
Avoid parsing LESS, PostCSS too

hmsk:
Other than Sass/SCSS, we may face parsing errors with `cssExtract.extractClasses(cssCode);` in `processStyle (stylePart, filePath)` if you give code which is invalid as pure CSS. `css` module can't parse.

like 

```css
.testPostcCSS {
  body {
    @media screen {
      background-color: red;
    }
  }
}
```

This syntax is allowed for almost CSS preprocessors. Anyway, some other syntaxes may make an exception with `css` module. So, I just added exts to ignore compilation.


eddyerburgh:
Thanks 😀

PR #59, opened 2018-02-20
fix display scss warning only once

### LINKED ISSUE #57: scss/sass warning
Currently the following message is displayed in the output every vue component containing scss or sass. 

`[vue-jest]: scss and sass are not currently compiled by vue-jest.`

I think that it can be displayed once at once.

What do you think?

km-tr:
This PR fixes #57


km-tr:
fix test

PR #61, opened 2018-02-21
mapCoverage removed from Jest

SimenB:
It's no longer needed - Jest handles it by default.

See https://github.com/facebook/jest/pull/5177


eddyerburgh:
Great, looks like I can close this—https://github.com/facebook/jest/issues/4509

Issue #69, opened 2018-03-15
Too many logs related to Sass


rmartins90:
https://cl.ly/3U3z3K2K1P3H

eddyerburgh:
Yeah this isn't ideal. 

We can either remove the warning, which could sting some developers, or add sass support.

rmartins90:
There's no way to emit only one warning?

eddyerburgh:
Yes actually it was easy to only log it once. I've added it in https://github.com/vuejs/vue-jest/commit/26e63066d9a09330ca0e82df8a0e6e0a3d59b698 . It will be in the next release. Thanks for bringing this to my attention!

ilyaztsv:
@eddyerburgh Hello!

As I can see [26e6306](https://github.com/vuejs/vue-jest/commit/26e63066d9a09330ca0e82df8a0e6e0a3d59b698) did not resolve the issue.

Please see the example repo: https://github.com/ilyaztsv/vue-jest-warnings

### Reproducing steps:

```
npm i
npm run test
```

ilyaztsv:
can be closed because https://github.com/vuejs/vue-jest/issues/75#issuecomment-379664086

PR #71, opened 2018-03-16
the errors message emits every file just one time

rchaser53:
related: https://github.com/vuejs/vue-jest/issues/69

the error messages are emit every import now.
it's a noisy for me.  so i send the PR to improve the error handling. 
i think vue-jest should emit error messages to notify the file name just one time.

### DIFF
--- a/lib/process.js
+++ b/lib/process.js
@@ -59,6 +59,7 @@ function changePartsIfFunctional (parts) {
   }
 }
 
+const BlackPathList = []
 module.exports = function (src, filePath) {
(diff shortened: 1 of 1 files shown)


rchaser53:
I have no idea to add a test for this PR.
do you have a good idea?

eddyerburgh:
Thanks for the PR! 

Stupidly I just added a commit that achieves the same thing before I saw this!

eddyerburgh:
> I have no idea to add a test for this PR.
> do you have a good idea?

No need to write a test for this. We just need tests that check the files are compiled correctly

PR #73, opened 2018-03-16
Scss integration

btoo:
hi @eddyerburgh - thanks for your awesome contributions to the vue ecosystem

i've been needing an scss compiler with this package for a minute now, so i thought i'd go ahead and try adding it myself. to give you an overview of this PR, i
- refactored a bit of the `processStyle` function and moved it to its own module
- added [node-sass](https://github.com/sass/node-sass) as a peer dependency
- added an `scss` compiler (no support for `sass` yet)
- allowed including `scss` global files via jest global (as recommended [here](https://github.com/vuejs/vue-jest/pull/53#discussion_r166065152)). the format of this feature is inspired by [sass-resources-loader](https://github.com/shakacode/sass-resources-loader)
- updated the README to reflect supported style languages

what do you think?


btoo:
you got it 👌

as for your question about rewriting imports, i've actually copied the functionality almost identically from [sass-resources-loader](https://github.com/shakacode/sass-resources-loader/blob/master/src/utils/rewriteImports.js#L26). what we're doing is converting all the sass import paths to absolute paths (starting at the project root) because `node-sass` is executing the compilation from wherever `scss-compiler.js` is located, which may change from user to user, rendering relative paths pretty risky. i believe that even if we do decide to keep relative paths in the mix, we'd still have to rewrite the imports such that they are relative to the files doing the importing rather than `scss-compiler.js` which i am sure is going to be a huge headache to do haha

Issue #63, opened 2018-02-27
Having both css modules and scss break test.


leRenart:
I've isolated an edge case when using css modules and scss generate a `Comment` where a `HTMLDivElement` is expected. 

**spec.js**

`const Constructor = Vue.extend(myComponent);
const vm = new Constructor().$mount();`

**myComponent**

This doesn't work

```
<template lang="pug">
  div( :class="$style.box" ) hello
</template>

<style module lang="scss">
.box {
  color: red;
}
</style>

```

This works

```
<template lang="pug">
  div( :class="$style.box" ) hello
</template>

<style module>
.box {
  color: red;
}
</style>
```

eddyerburgh:
This should be fixed by #73

I've release in 2.2.0. Can you confirm the issue has been fixed?

jeremyzahner:
@leRenart @eddyerburgh Had the same issue. Found some of the (alleged) problems at least. Also added a unit test that should help fixing this. See https://github.com/vuejs/vue-jest/pull/79

PR #79, opened 2018-04-09
WIP: Fixes styles when combining modules and SCSS

jeremyzahner:
I've had the problem (https://github.com/vuejs/vue-jest/issues/63) that either modules or SCSS would work. They don't seem to do some when combined.

I tried to figure out what the problem is. So far, i've found an issue in the SCSS compiler which i believe i fixed. Still, the classes seem not to end up on the $style of the final VM.

Important: This is WIP and not ready to be merged. I guess it can be if the test suite runs through.


jeremyzahner:
Ping @danielbastos11 @btoo

jeremyzahner:
@eddyerburgh It should be fixed now.

jeremyzahner:
@eddyerburgh Found some more bugs. Fixed them and added new tests. New PR incoming. (Just to let you know so you don't release something thats not quite working.)

PR #93, opened 2018-06-06
Improve .babelrc lookup

TheCycoONE:
Now searches with a depth of infinity from the file path, as described
in #66.

For my personal use, just switching the depth from 0 to the default (INFINITY) was enough; but as per the ticket I try to use the filePath now as well.

### DIFF
--- a/lib/compilers/babel-compiler.js
+++ b/lib/compilers/babel-compiler.js
@@ -1,8 +1,8 @@
 const babel = require('babel-core')
 const loadBabelConfig = require('../load-babel-config.js')
 
-module.exports = function compileBabel (scriptContent, inputSourceMap, inlineConfig, vueJestConfig) {
-  const babelConfig = inlineConfig || loadBabelConfig(vueJestConfig)
+module.exports = function compileBabel (scriptContent, inputSourceMap, inlineConfig, vueJestConfig, filePath) {
+  const babelConfig = inlineConfig || loadBabelConfig(vueJestConfig, filePath)
 
   if (!babelConfig) {
     return {
--- a/lib/compilers/coffee-compiler.js
+++ b/lib/compilers/coffee-compiler.js
@@ -2,7 +2,7 @@ var ensureRequire = require('../ensure-require.js')
 const throwError = require('../throw-error')
 const loadBabelConfig = require('../load-babel-config.js')
 
-module.exports = function (raw, vueJestConfig) {
+module.exports = function (raw, vueJestConfig, filePath) {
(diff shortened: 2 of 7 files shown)


TheCycoONE:
The issue is that the test is being run in `vue-jest/temp-dir` where the .babelrc is being renamed; but the search finds the babelrc in vue-jest now.

eddyerburgh:
Thanks for the PR. 

What's your use case?

TheCycoONE:
The ability to run jest from anywhere in the npm project, particularly the location of the spec file.

In particular this allows IntelliJ's Jest plugin to run without configuration in the IDE. It is also consistent with babel-jest.

eddyerburgh:
We should keep the same functionality as babel-jest, but I don't think babel-jest searches in parent directories for the babel config file. 

I'll have to take a deeper look at the source code. From what I could tell the babel options are found when the transformer is initially created, which I think is in the project root. We could implement something similar.

TheCycoONE:
My understanding is that this loop removes the last part of the path until it hits the root or escapes due to cache or finding the config. I have made an assumption about the filename being passed in, I could test that later if you like. https://github.com/facebook/jest/blob/4a46993b5116838a55666ac8c0522a71457eb7c5/packages/babel-jest/src/index.js#L37

eddyerburgh:
You're right, it does.

Ok, how about we rename the root babelrc as part of the test script, so it doesn't find it when the tests are running, then rename when the process exits?

TheCycoONE:
My feeling is that it is better to mock our findBabelConfig dependency so we aren't escaping the test sandbox, and the unit tests are more contained. I'm quite open to your feedback though

TheCycoONE:
I made the test changes as you suggested.

eddyerburgh:
Hey, sorry that I've been silent on this. It slipped my radar.

I agree that we should add this feature, but we need a test for the filepath functionality specifically. Are you able to add a test to check that the babelrc searching uses the current directory?

TheCycoONE:
Sure, no problem. Likely tomorrow.

TheCycoONE:
@eddyerburgh Are those the tests you wanted?

eddyerburgh:
Thanks for adding the tests. I was thinking they should be tests that call through the whole `process` function, like [these tests](https://github.com/vuejs/vue-jest/blob/master/test/Babel.spec.js#L37). The reason is that we will be refactoring the project soon and the load-babel-config function might be removed/ refactored. But they are fine as is. Thanks a lot for your patience and for creating this PR :)

PR #100, opened 2018-08-17
Allow babelInlineConfig to be passed in

Globegitter:
We are creating our babelConfig programatically and with babel-jest it is already possible to pass that in, but not here. With these changes I can now write my custom vue-jest transform in the following way:

```
...
module.exports = {
    process(src, filename, jestConfig) {
        return vueJest.process(src, filename, jestConfig, {}, babelOptions);
    },
};
```

### DIFF
--- a/lib/load-babel-config.js
+++ b/lib/load-babel-config.js
@@ -14,7 +14,9 @@ module.exports = function getBabelConfig (vueJestConfig) {
     let babelConfig
 
     if (vueJestConfig.babelRcFile) {
-      babelConfig = require(vueJestConfig.babelRcFile)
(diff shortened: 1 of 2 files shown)


Globegitter:
Tests are passing now.

eddyerburgh:
I think this should be added to the Jest config option object, rather than added as an extra object. Could you also please add some tests?

justinhelmer:
@eddyerburgh - done. Check #110 / #111 .

justinhelmer:
I believe this PR can be closed now that #111 has been merged.

Globegitter:
@justinhelmer thanks for that I will check it out but it does sound good.

PR #104, opened 2018-09-04
Remove yarn.lock file

rimiti:
Hello everybody,

Using two locker (`yarn.lock` and `package-lock.json`) files isn't usefull. This PR only remove the `yarn.lock` file.

Regards, 🚀


rimiti:
This PR is WIP but I do not have time to finish it.

eddyerburgh:
Yes we should only have 1 lock file, the convention in other Vue repos is to use yarn, rather than npm

rimiti:
Thank you @eddyerburgh for your reply. I'll update this PR to remove package-lock.json file.

PR #111, opened 2018-10-03
feat: update to support babelConfig and tsConfig

### LINKED ISSUE #110: can't use any programmatic babel options
## The problem
Currently, this tool supports supplying a custom `.babelrc` file location via Jest global configuration (`jest.globals['vue-jest'].babelRcFile`), which is great.

However, `.babelrc` is not an option for us. Due to the fact that it is JSON, we cannot have any logic in the babel configuration. I tried using `.babelrc.js`, which `babel` supports, but the config loader assumes JSON and [parses the file](https://github.com/vuejs/vue-jest/blob/d9f933641f2cb1e5e6e565a8f311431809732526/lib/load-babel-config.js#L17).

Additionally, we have moved to the new [project-wide configuration](https://babeljs.io/docs/en/config-files#project-wide-configuration) file (`babel.config.js`) which is [recommended by babel](https://babeljs.io/docs/en/configuration) for our use-case, and exports a function instead of an object.

## The solution
expand the global jest configuration for `vue-jest` to support arbitrary babel options:

```json
{
  "jest": {
    "globals": {
      "vue-jest": {
        "babelOptions": {
          "presets": [
            [
              "env",
              {
                "useBuiltIns": "entry",
                "shippedProposals": true
              }
            ]
          ],
          "plugins": [
            "syntax-dynamic-import"
          ],
          "env": {
            "test": {
              "plugins": [
                "dynamic-import-node"
              ]
            }
          }
        }
      }
    }
  }
}
```

This generic solution opens up many more possibilities for how the tool can be used. For example, if the [--config](https://jestjs.io/docs/en/cli.html#config-path) option is used with Jest, we are no longer restricted to JSON. This allows for the configuration to have logic:

```js
// @file jest.config.js
const path = require('path');
const config = require('./config');

const envPreset = [
  '@babel/preset-env',
];

if (config.foo) {
  envPreset.push({
    useBuiltIns: 'entry',
    shippedProposals: true,
  });
}

module.exports = {
  globals: {
    'vue-jest': {
      babelOptions: {
        presets: [envPreset],
        plugins: ['babel-plugin-*'],
      }
    }
  }
}
```

It can also be used to support the [configFile](https://babeljs.io/docs/en/options#configfile) option to use a [project-wide configuration](https://babeljs.io/docs/en/config-files#project-wide-configuration) file (`babel.config.js`), e.g.:

```js
// @file jest.config.js
module.exports = {
  globals: {
    'vue-jest': {
      babelOptions: {
        configFile: path.resolve(__dirname, './babel.config.js'),
      }
    }
  }
};
```

This configuration file exports a function that can use the [config function API](https://babeljs.io/docs/en/config-files#config-function-api) and returns babel options:

```js
// @file babel.config.js
module.exports = (api) => {
  api.cache(() => process.env.NODE_ENV === 'production');
  return {
    presets,
    plugins
  };
};
```

This is especially powerful when used in conjunction with `webpack`, which can supply the same configuration to it's `babel-loader` to ensure functional parity when building the application.

This solution also allows for consumers to take advantage of the new [configuration file resolution](https://babeljs.io/docs/en/next/config-files) logic built-in to babel by using the `root` option:

```js
// @file jest.config.js
module.exports = {
  globals: {
    'vue-jest': {
      babelOptions: {
        root: path.resolve(__dirname),
      }
    }
  }
};
```

Although I have not explored this option, it could provide an alternative approach to circumvent the current logic provided by third-party [find-babel-contrib](https://github.com/tleunen/find-babel-config), which is not owned/maintained by the babel core team. I also found an open PR (#93) which indicates there are needs to make it better, or closer to the native behavior of `babel`.

justinhelmer:
Fixes #110

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 Jest Vue transformer with source map support
 
+> **NOTE:** This is documentation for `vue-jest@3.x`. [View the vue-jest@2.x documentation](https://github.com/vuejs/vue-jest/tree/e694fc7ce11ae1ac1c778ed7c4402515c5f0d5aa)
+
 ## Usage
 
 ```bash
@@ -10,7 +12,7 @@ npm install --save-dev vue-jest
 
 ## Setup
 
-To define vue-jest as a transformer for your .vue files, you need to map .vue files to the vue-jest module.
+To define `vue-jest` as a transformer for your `.vue` files, map them to the `vue-jest` module:
 
 ```json
 {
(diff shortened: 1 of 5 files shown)


justinhelmer:
Changed I needed locally to contribute: #108 
Replaces the need for: #100 
Potential alternative approach for: #93

justinhelmer:
You may notice the configuration option is called is `babelOptions` instead of `babelConfig` (which is used internally by `vue-jest`). I originally had called it `babelConfig`, but changed it.

I decided this because other community tools use the nomenclature `babelOptions`, including [babel itself](https://babeljs.io/docs/en/options).

justinhelmer:
@eddyerburgh - To match the [ts-jest API](https://kulshekhar.github.io/ts-jest/user/config/babelConfig), it would result in a breaking change to the current default logic for `vue-jest`. 

From the docs:

> The option is babelConfig and it works pretty much as the tsConfig option, except that it is disabled by default. Here is the possible values it can take:

> `false`: the default, disables the use of Babel
> `true`: enables Babel processing. ts-jest will try to find a .babelrc, .babelrc.js file or a babel section in the package.json file of your project and use it as the config to pass to babel-jest processor.
> `{ ... }`: inline Babel options. You can also set this to an empty object ({}) so that the default Babel config file is not used.

Currently, even if *no* configuration is provided, the default [fallback logic](https://github.com/vuejs/vue-jest/blob/master/lib/load-babel-config.js#L21) for `vue-jest` is to search for babel config. This would equate to `babelConfig: true` as the default behavior given the changes.

So the question is,

1) Do you want me to proceed with matching the API exactly, including default behavior? This may need to result in a major version bump. Not sure if you want to go down this route.

2) Do you want me to match the API with the *one* difference being the default behavior of `true` vs `false`?

3) What about the "option as a string" that assumes a path, per your request in the comment above? `ts-jest` doesn't have this functionality. Do you want me to add this on top of their API?

As a reminder, as I mention in the issue related to this PR (#110), babel now natively supports passing a configuration file directly:

```json
{
  "babelConfig": {
    "configFile": "path/to/file"
  }
}
```

Also, can you clarify what you mean by "not just unit tests"? It looks to me like [that's all there is](https://github.com/vuejs/vue-jest/blob/master/package.json#L26-L28)

eddyerburgh:
> Do you want me to proceed with matching the API exactly, including default behavior? This may need to result in a major version bump. Not sure if you want to go down this route.

1. Not exactly, we should default to true

2. Do you want me to match the API with the one difference being the default behavior of true vs false?

Yes

3. What about the current babelRcFile config? If babelConfig also supports a string representation as a path to a babel config file, then babelRcFile would be repetitive. Are you suggesting removing this option as well or leaving it as-is? If removing it, that would definitely dictate a major version bump.

I think we should deprecate the option as part of this work. To deprecate it, we need to remove it from the docs and add a warning when the method is called (without removing functionality).

I also think we should keep the API the same between `tsConfig` and `babelConfig`. Could you also deprecate the `tsConfigFile` option, and add a new `tsConfig` file option that behaves the same way?

Sorry for the amount of work, but you highlighted an issue with our globals design and I think we should address it now. Also, the next release will be a major due to [this PR](https://github.com/vuejs/vue-jest/pull/93) which changes the  behavior of the babelrc lookup

justinhelmer:
Sure, no worries. I was coming to the same conclusion.

Two follow-up questions: 

1) We are flipping the default behavior of `babelConfig` when compared against `ts-jest`. In other words, `vue-jest` will attempt to search for config and apply Babel transforms unless `babelConfig` is explicitly set to `false`. 

Are we flipping the default behavior of `tsConfig` as well? Or will the `tsConfig` API match [ts-jest](https://kulshekhar.github.io/ts-jest/user/config/tsConfig) *exactly*, including the default behavior?

If so, then `vue-jest` will search for both ts config and babel config (and transpile automatically) unless explicitly disabled.

2) Is it safe to assume you will be handling the versioning and I can leave that part alone? i.e. merging with master won't publish a minor version bump through some automated system, and break current consumers?

justinhelmer:
Requested changes in place.

Build will fail, because there is a prerequisite on #113

eddyerburgh:
Hey @justinhelmer I decided that we should just let ts-jest and babel-jest handle compiling, like you suggested. Would you like to work on ts-jest integration?

justinhelmer:
@eddyerburgh - I believe you are referring to #114 .

Unfortunately I don't have the bandwidth right now to take on this challenge. It might change in the future if someone else doesn't take this on, but I don't want to be a bottleneck for delivery.

PR #113, opened 2018-10-05
feat: add deprecation message lib

justinhelmer:
Prerequisite for #111 (deprecating `babelRcFile`, `tsConfigFile`)

### DIFF
--- a/lib/deprecate.js
+++ b/lib/deprecate.js
@@ -1,0 +1,12 @@
+const chalk = require('chalk')
+const logger = require('./logger')
+
+const replace = (oldThing, newThing) => {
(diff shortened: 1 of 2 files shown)


justinhelmer:
Apologies for not spotting that. Squashed and pushed.

justinhelmer:
One thing to mention - I was noticing the output was a bit funky and inconsistent. I tried playing with `process.stdout.write` but no luck. Nothing aside from logging the same thing multiple times would ensure the warning would be logged to completion.

I did some digging and found https://github.com/facebook/jest/issues/3853 and https://github.com/facebook/jest/issues/2441 so I concluded there wasn't much we can do until those are resolved.

PR #121, opened 2018-10-25
Fix repo url in package.json

ankurk91:
The repo has been moved under vuejs organisation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,6 @@
   },
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/eddyerburgh/vue-jest.git"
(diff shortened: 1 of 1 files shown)


jackkoppa:
whooo! Is there a reason that the NPM page is still linking to @eddyerburgh's repo? https://github.com/eddyerburgh/vue-jest

Screen Shot 2019-04-04 at 12 07 11 PM

It looks like the `v3` branch still list's that original repo; not sure if that's where the most recent publishes have been done from?

https://github.com/vuejs/vue-jest/blob/v3/package.json#L88-L91

eddyerburgh:
Yes good point, would you like to update the v3 branch and I can publish it?

PR #131, opened 2018-11-24
use correct source map names for script files imported over src

steplov:
Hello. This PR fixes the issue when Idea IDE uses *.vue file instead of script file included with <script src=""
Chrome Dev Tools still works well. Would be nice if someone can confirm that VSCode works correct as well


steplov:
@eddyerburgh hello, could you please take a look into the issue with circle CI. Looks like the issue with build pipeline itself

steplov:
@eddyerburgh thanks a lot

Issue #122, opened 2018-11-03
<script lang="ts"> not working with babel 7, @babel/preset-typescript


rumanbsl:
Hi guys, I am using babel 7 and @babel/preset-typescript. The js scripts parses, and I do not have any issues, but in `<script lang="ts">` it is throwing error:

![image](https://user-images.githubusercontent.com/13895272/47952376-13bc2600-df77-11e8-93f1-10ea47bccbc4.png)

here is my package.json file
```js
{
  "name": "typescript-project",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "dev": "cross-env NODE_ENV=dev webpack",
    "prod": "rimraf dist && cross-env NODE_ENV=prod webpack",
    "start": "cross-env NODE_ENV=dev webpack-dev-server",
    "test:unit:debug": "node --inspect-brk ./node_modules/jest/bin/jest.js --no-cache --runInBand",
    "test": "jest"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "@babel/cli": "7.0.0-beta.42",
    "@babel/core": "7.0.0-beta.42",
    "@babel/node": "7.0.0-beta.42",
    "@babel/plugin-proposal-class-properties": "^7.0.0-beta.42",
    "@babel/preset-env": "7.0.0-beta.42",
    "@babel/preset-react": "7.0.0-beta.42",
    "@babel/preset-stage-0": "7.0.0-beta.42",
    "@babel/preset-typescript": "7.0.0-beta.42",
    "@babel/register": "7.0.0-beta.42",
    "@kazupon/vue-i18n-loader": "^0.3.0",
    "@mapbox/stylelint-processor-arbitrary-tags": "^0.2.0",
    "@types/jest": "^23.3.9",
    "@types/lodash": "^4.14.104",
    "@types/node": "^10.12.2",
    "@types/vue-router": "^2.0.0",
    "@types/webpack-env": "^1.13.6",
    "@vue/test-utils": "^1.0.0-beta.25",
    "autoprefixer": "^9.1.3",
    "babel-core": "^7.0.0-bridge.0",
    "babel-jest": "^23.6.0",
    "babel-loader": "8.0.4",
    "caniuse-lite": "^1.0.30000903",
    "cross-env": "^5.1.3",
    "css-loader": "^1.0.1",
    "cssnano": "^4.1.0",
    "eslint": "^5.2.0",
[...]

PR #138, opened 2018-12-26
Babel bug: `.inputSourceMap must be a boolean, object, or undefined` & `filename` is undefined error fix

rumanbsl:
This is my first pull request ever just FYI :) So, Please go easy on me.

Current @babel/core requires either boolean, or object in inputSourceMap. So putting empty string will throw error (lib/compilers/typescript-compiler.js : 14)

While using @babel/preset-typescript babel.transform requires a filename(lib/compilers/babel-compiler.js: 20), even though the typedef for @babel/core says it's optional. I have given the defaultname "unknown" mentioned in the @types/babel-core file

Ran test to make sure it didn't break anything


rumanbsl:
This is my first pull request ever just FYI :)

rumanbsl:
Can you please elaborate the steps? As in how/what should I test?

eddyerburgh:
You could create a test that reproduces the issue.

You can see a test here—https://github.com/vuejs/vue-jest/blob/v3/test/Babel.spec.js#L98. It writes plugins to the .babelrc file, then calls `jestVue.process`.  In your case, you would add @babel/preset-typescript to the .babelrc, and then make sure it doesn't error by calling `vueJest.process` (assuming that it does at the moment). Let me know if you have any other questions

rumanbsl:
Been banging my head since yesterday. Currently `@babel/core` is required for `@babel/preset-typescript`, it is throwing bunch of errors when I install @babel/core in 
image

no luck on testing :(

lib/compilers/babel-compiler.js
```diff
@@ -1,4 +1,4 @@
-const babel = require('babel-core')
+const babel = require('@babel/core')
```
eddyerburgh: This makes this PR a breaking change, can you keep the old babel-core in this PR? You can use babel-core bridge (https://github.com/babel/babel-bridge) if you need functionality of babel 7.

rumanbsl:
Alright, so it was not as easy I thought it would be, because it is to do with babel 7, and the current vue-jest is transpiled with babel 6. Here are the summary of what has been done:

1. Babel is upgraded to version 7
2. All the related code is upgraded to comply with babel changes.
3. A test case is setup, where the issue is reproduced.

As `babel-plugin-transform-es2015-modules-commonjs` is replaced with `@babel/plugin-transform-modules-commonjs` the snapshot test is breaking. Please take steps, as I am not familiar with snapshot test(haven't read the chapter from your vue testing book yet 😅 ) Also please bump up the node version, as Yarn is failing in circleci if I upgrade the eslints as will (optional)

rumanbsl:
#122 should be fixed with this one, along with some other babel 7 related issues which are faced by users

lib/compilers/babel-compiler.js
rumanbsl: Sorry, I didn't notice this earlier. Would it be possible for you to review the latest code, as I have changed codebase heavily. I will revert back to babel 6 if you think that's a must

lib/compilers/babel-compiler.js
```diff
@@ -23,7 +23,16 @@ module.exports = function compileBabel (scriptContent, inputSourceMap, inlineCon
     babelOptions = Object.assign(babelOptions, babelOverWrites)
   }
 
-  const res = babel.transform(scriptContent, babelOptions)
```
eddyerburgh: Can you revert to the previous code? babel.transform will use babel 6 transform if babel-core is babel 6, and babel 7 if babel-core is @bridge.

rumanbsl: My tests were failing when i reveretd back :/
image

rumanbsl: I reverted back to original code, except this chunk
image
This to only reproduce the issue in test

Now the snapshot test is failing, can you check the codes for assesment

rumanbsl:
@eddyerburgh Test provided, snapshot failing. Please check, My test included babel 7, which is why there is snapshot missmatch.

CinKon:
For anyone who encounters this error too: After several hours I found a fix/workaround for this problem:

tsconfig.json:
```json
{
  "compilerOptions": {
    [...]
    "allowSyntheticDefaultImports": true,
```

This is the important setting.

kermorgant:
Same error happened to me, and in my case, I solved it by adding `"sourceMap": true` in tsconfig.json

lib/process.js
```diff
@@ -27,19 +27,19 @@ function processScript (scriptPart, vueJestConfig, filePath) {
...
   const vueJestConfig = getVueJestConfig(jestConfig)
 
   var parts = vueCompiler.parseComponent(src, { pad: true })
 
   if (parts.script && parts.script.src) {
     parts.script.content = fs.readFileSync(join(filePath, '..', parts.script.src), 'utf8')
   }
-
-  const result = processScript(parts.script, vueJestConfig, filePath)
+  const result = processScript(parts.script, vueJestConfig, filePath, babelTestObj)
```
eddyerburgh: Can you please remove the `babelTestObj` from the code

test/Babel-typescript.spec.js
```diff
@@ -0,0 +1,62 @@
...
+} from 'fs'
+
+import clearModule from 'clear-module'
+import cache from '../lib/cache'
+
+beforeEach(() => {
+  cache.flushAll()
+  clearModule.all()
+})
+test('babel 7 configuration expects filename [default: "unknown"] to compile', () => {
```
eddyerburgh: In this case, since it's a small change and difficult to test, you can remove these tests and we'll leave it untested

eddyerburgh:
So since the changes to code are relatively small, and we'll realase v4 soon, it's OK to add this code without any tests. Can you please put an inline comment above the code to add the filename, and above the ternary where you assign the input map, to describe why you made the changes?

lib/process.js
rumanbsl: done

test/Babel-typescript.spec.js
rumanbsl: done

lib/compilers/babel-compiler.js
```diff
@@ -17,11 +17,15 @@ module.exports = function compileBabel (scriptContent, inputSourceMap, inlineCon
   }
 
   const babelOptions = Object.assign(sourceMapOptions, babelConfig)
+  if (!babelOptions.filename) babelOptions.filename = 'unknown'
```
eddyerburgh: Can you remove this line please

eddyerburgh:
This has been released in 3.0.3 👍

mattgrande:
@eddyerburgh - I'm not seeing 3.0.3 on NPM yet. Is there a delay on NPM's side?

eddyerburgh:
No sorry, the publish had failed on my end. Just published now

Issue #125, opened 2018-11-12
Feature Request: Add support for CSS Modules localIdentName


SimulatedGREG:
When using CSS modules in combination with `@vue/test-utils`, the DOM is created with class names equal to those that are defined within the `<style/>` block. For my company we use a custom `localIdentName` that is different from this default. It would be nice if the outputted class names could use a custom [`localIdentName`](https://github.com/webpack-contrib/css-loader#localidentname).

This is useful for us specifically because we take an approach where each UI component we write should be wrapped with a `.component` class. So with a `localIndentName` of `[name]-[local]` each component would output something with a class name like `.ui-button-component` or `.ui-input-component` (where `ui-button` and `ui-input` are the filenames of our components).

Without the support for `localIdentName` in `vue-jest`, all of our wrapper classes are `.component` which, I'm sure you can imagine, can make it difficult when testing UI states where a specific DOM lookup is needed.

Hopefully this is enough to get my idea across, if not please feel free to ask any questions or perhaps I might be able to provide a thorough example. I am also aware since there isn't any use of `webpack`, nor more specifically `css-loader`, this may not be possible. But of course, please share any ideas for getting this to work!

PR #146, opened 2019-01-10
add pre & post css transform callbacks to customize css transformatio…

### LINKED ISSUE #145: Support for pre/post transformers similar to webpack loaders
I'm working on a project which builds the app for multiple themes and enables user to switch between them on the fly. We use mixins to set `$selectedTheme = this.$style[theme]` on the component. In the SFC, we do as below:

```
<template>
    <div :class="$selectedTheme['active']">
        Content
    </div>
</template>

<style lang="scss" module themed>
   .active{
      @include active();
   }
</style>
```
We use webpack custom loaders to identify the `themed `attr and build css modules for all the themes.
When we try to test such components, we get error as we are not transforming those css modules.

I know we can mock $styles while we are mounting the components but I believe a **pre & post transform hook** in vue-jest for every type is a **good to have feature** for supporting features  webpack loaders do.

I'm happy to open a PR if it is a valid feature.

yohodopo:
pre & post css transform callbacks to customize css transformation behavior 
Fixes #145,  #125 by allowing users to write custom transformers similar to webpack loaders


vamc-anne:
Hey @eddyerburgh , I've updated the PR with test cases, can you please look into it. Thanks

lib/compilers/helpers/module-name-mapper-helper.js
```diff
@@ -1,17 +1,22 @@
 const path = require('path')
-
+const matchModuleImport = /^[^?]*~/
 /**
- * Resolves the path to the file locally.
+ * Resolves the path to the file/modeule.
```
eddyerburgh:
```suggestion
 * Resolves the path to the file/module.
```

lib/compilers/helpers/module-name-mapper-helper.js
```diff
@@ -1,17 +1,22 @@
...
  *
  * @param {String} to - the name of the file to resolve to
- * @param {String} localPath - the local path
+ * @param {String} importPath - the local path
+ * @param {String} fileType - extn of the file to be resolved
```
eddyerburgh:
```suggestion
 * @param {String} fileType - extension of the file to be resolved
```

eddyerburgh:
I agree with this PR, but I think we should make it more general. 

The config option should be `transform`:

```js
{
   jest: { 
    'vue-jest': {
      transform: {
        scss: './custom-transform'
      }
    }
  }
}
```

And each transform should expect to be the same as any Jest transformer:

```js
// custom-transform.js
module.exports = {
  process(src, filePath, config) {
    // custom transform
  }
}
```

We could support a `preProcess` and `postProcess` script that receives the same parameters, to support the functionality you're proposing:

```js
// custom-transform.js
module.exports = {
  process(src, filePath, config) {
    // custom transform
  },
  postProcess(src, filePath, config) {
    // ..
  },
  preProcess(src, filePath, config) {
    // ..
  },
}
```

So in this PR can you make changes to the API to use `transform` as the config option, and expect the transforms to be an object that can contain postProcess and preProcess methods.

yohodopo:
@eddyerburgh, I've updated the PR to make the change generic.

eddyerburgh:
Thanks for the changes, I'm going to give this a full review at the weekend

Issue #148, opened 2019-01-17
Can't resolve path to tsconfig when using tsConfig


Loremaster:
Hello!

I upgraded `vue-jest` from 2.6.0 to 3.0.2 and it started to show deprecation message:

```
[vue-jest]: Deprecation Warning:

[vue-jest]: Option "tsConfigFile" has been removed, and replaced by "tsConfig"

[vue-jest]: This option will be removed in the next major version of `vue-jest`. Please update your configuration.
```

so I replaced `tsConfigFile` with `tsConfig` but it doesn't seem to resolve path properly now for some reasons:

```
 FAIL  app/javascript/compose_email/components/Emails/PersonalText/__tests__/Panel.spec.ts
  ● Test suite failed to run

    Cannot find module 'tsconfig.jest.json'

      at loadTypescriptConfig (node_modules/vue-jest/lib/load-typescript-config.js:53:30)
      at compileTypescript (node_modules/vue-jest/lib/compilers/typescript-compiler.js:9:20)
      at processScript (node_modules/vue-jest/lib/process.js:23:12)
      at Object.module.exports [as process] (node_modules/vue-jest/lib/process.js:42:18)
```

Can you give an advice how that can be fixed?

**package.json**
```
{
  "engines": {
    "node": "11.6.0",
    "npm": "5.6.0",
    "yarn": "1.12.3"
  },
  "dependencies": {
    "@rails/webpacker": "^3.5.5",
    "@types/ckeditor": "^4.9.5",
    "@types/jquery": "^3.3.29",
    "@types/lodash": "^4.14.119",
    "@types/node": "^10.12.18",
    "@types/pusher-js": "^4.2.0",
    "axios": "^0.18.0",
    "babel-loader": "^7.1.5",
    "babel-polyfill": "^6.26.0",
    "babel-preset-vue": "^2.0.2",
    "bootstrap-vue": "^2.0.0-rc.11",
    "coffeelint": "^2.1.0",
    "epic-spinners": "^1.0.4",
    "es6-promise": "^4.2.5",
    "immutable": "^4.0.0-rc.12",
    "jquery": "^3.3.1",
    "lodash": "^4.17.11",
    "nodelist-foreach-polyfill": "^1.2.0",
    "pusher-js": "^4.3.1",
    "qs": "^6.6.0",
    "rosie": "^2.0.1",
    "ts-loader": "3.5.0",
    "typescript": "^3.2.2",
    "vee-validate": "^2.1.5",
[...]

Issue #144, opened 2019-01-09
V4.0.0-beta.0: extends in tsconfig.json does not work as expected.


vamc-anne:
tsConfig.json
```
{
  "extends": "./.temp/tsconfig-base.json"
}
```
tsconfig-base.json
```

{
  "include": ["../src/**/*", "../types/**/*.d.ts", "./loader-types.d.ts", "./client-types.d.ts"],
  "exclude": ["../src/**/*.spec.ts", "../src/**/*.spec.js"],
  "compilerOptions": {
    "strict": true,
    "skipLibCheck": true,
    "module": "esnext",
    "target": "es6",
    "allowSyntheticDefaultImports": true,
    "moduleResolution": "node",
    "lib": ["dom", "es6", "dom.iterable", "scripthost"],
    "noEmitHelpers": true,
    "importHelpers": true,
    "inlineSources": true,
    "sourceMap": true,
    "experimentalDecorators": true,
    "emitDecoratorMetadata": true,
    "baseUrl": ".",
    "paths": {
      "~ngAppModule": ["./ng-app-module"]
    },
    "outDir": "../dist/",
    "rootDir": "../src"
  }
}
```
When I tried debugging this issue, it logs 
`{"extends":"./.arrow/tsconfig-base.json","compilerOptions":{}}` at https://github.com/vuejs/vue-jest/blob/master/lib/compilers/typescript-compiler.js#L11

related to #118

ibuchan72390:
I can confirm this issue as well.  I have been attempting to implement testing in Nuxt leveraging Jest.  The default recommended config for nuxt-ts is:

```
{
  "extends": "./node_modules/nuxt-ts/tsconfig.nuxt-ts.json"
}
```

Eventually I chased it down to a failure to load the extended attributes into the referenced object as previously mentioned by @yohodopo

FeliciousX:
can confirm this is happening aswell :(

i was trying to debug why my [typescript decorator](https://www.typescriptlang.org/docs/handbook/decorators.html) is having it's 3rd argument as `undefined` in the test but has value in app code.

> NOTE The Property Descriptor will be undefined if your script target is less than **ES5.**

I assumed it was in `vue-jest` because it was trying to test a `.vue` file with `lang="ts"`

I tried tweaking the `tsConfig` option for `vue-jest` and finally managed to reproduce it if i didn't extend my config.. which is where my `target` option lies

IlCallo:
[...]

PR #154, opened 2019-01-30
let the script processer pick custom transformers when present

yohodopo:
This PR enables the use of custom script transformers, resolves extends property from tsconfig.json.

Similar to custom CSS transformers this help when we have any custom processing to the script(ts/js/coffee).

For eg. When users do not have babel.config.js at the root of the application but have it at a specific location, we can create a custom babel-transformer by passing in babel-jest's `createTransformer(babelConfig)`  with the config pointing to the babel configuration.

Fixes, #148, #144

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -64,7 +64,8 @@ const getBabelOptions = function loadBabelOptions(filename, options = {}) {
 
 const getTsJestConfig = function getTsJestConfig(config) {
   const tr = createTransformer()
-  return tr.configsFor(config)
+  const { typescript } = tr.configsFor(config)
+  return { compilerOptions: typescript.options }
 }
 
 function isValidTransformer(transformer) {
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -93,7 +93,7 @@ const getCustomTransformer = function getCustomTransformer(
     }
     return transformer
   }
-  return {}
+  return null
 }
 
 const throwError = function error(msg) {
--- a/e2e/__projects__/custom-transformers/babel-transformer.js
+++ b/e2e/__projects__/custom-transformers/babel-transformer.js
@@ -1,0 +1,4 @@
+const { createTransformer } = require('babel-jest')
+module.exports = createTransformer({
+  presets: ['@babel/preset-env']
+})
--- a/e2e/__projects__/custom-transformers/babel.config.js
+++ b/e2e/__projects__/custom-transformers/babel.config.js
@@ -1,3 +1,0 @@
-module.exports = {
-  presets: ['@babel/preset-env']
-}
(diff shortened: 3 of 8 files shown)


lib/utils.js
```diff
@@ -64,7 +64,8 @@ const getBabelOptions = function loadBabelOptions(filename, options = {}) {
 
 const getTsJestConfig = function getTsJestConfig(config) {
   const tr = createTransformer()
-  return tr.configsFor(config)
+  const { typescript } = tr.configsFor(config)
+  return { compilerOptions: typescript.options }
```
yohodopo: Or we could do something like the this to make sure we dont depend on `ts-jest` to fetch tsConfig options when users do not have ts-jest as their configuration.

```
const getTsJestConfig = function getTsJestConfig(config = {}) {
  const vueTsConfig = getVueJestConfig(config)
  if (typeof vueTsConfig.tsConfig === 'string') {
    return getTsConfigJson(vueTsConfig.tsConfig)
  } else if (typeof vueTsConfig.tsConfig === 'object') {
    return vueTsConfig.tsConfig
  } else if (config.globals && config.globals['ts-jest']) {
    const tr = createTransformer()
    const {
      typescript: { options: compilerOptions }
    } = tr.configsFor(config)
    return { compilerOptions }
  } else {
    // auto pick tsconfig from the root folder
    return getTsConfigJson()
  }
}

const getTsConfigJson = function getTsConfigJson(
  pathToConfig = path.resolve(process.cwd(), './tsconfig.json')
) {
  const configJson = ts.parseConfigFileTextToJson(
    pathToConfig,
    ts.sys.readFile(pathToConfig)
  )
  const { options: compilerOptions } = ts.parseJsonConfigFileContent(
    configJson.config,
    ts.sys,
    ts.getDirectoryPath(pathToConfig),
    {},
    pathToConfig
  )
  return { compilerOptions }
}
```

yohodopo: @eddyerburgh thoughts?

eddyerburgh: For most use cases it would be fine to compile with Babel, it would also simplify the code. The only problem would be if users use const enums, namespaces, or `export =` or `import =`. 

I haven't seen anyone using those features in components, so I think we can go ahead with using Babel for compiling TypeScript.

eddyerburgh: Since we added ts-jest, Jest added native support for TypeScript in v24 (which uses Babel), so I think we are fine to remove ts-jest and just use Babel.

eddyerburgh: Can you please make this change in a future PR

eddyerburgh:
Are you planning to open a PR to remove TypeScript and use babel instead?

yohodopo:
Hey @eddyerburgh , I've not had enough time to work on it. Will create a PR over the weekend

Issue #76, opened 2018-03-29
Using external css


hewIngram:
Hello, 

I'm looking at adding unit test coverage to my vue app but I'm having trouble fetching my css. I'm currently using external css in my components:

``` <style module scoped src='./Example.css' /> ```

Which isn't working, I spotted @danielbastos11 's PR adding support for css modules and was wondering if vue-jest can deal with external css yet? If not I'm happy to extend the work in https://github.com/vuejs/vue-jest/pull/50 to allow external css files and open a PR for it.

Thanks

**Comment by eddyerburgh** [comment]
We  don't support  external CSS. A PR would be very welcome!

**Comment by rmartins90** [comment]
does this help @hewIngram? https://facebook.github.io/jest/docs/en/webpack.html#handling-static-assets

**Comment by lancetharper** [comment]
I think I've run into a similar (or might be the same) issue.

```
<style module>
@import "some/file.css"
</style>
```
`wrapper.vm.$style` = {}.

If I change it to:
```
<style module>
@import "some/file.css"

.test {
  color: purple;
}
</style>
```
`wrapper.vm.$style` = { test: 'test' }

Does this look to be the same issue?

hewIngram:
Just opened a PR that reads the src attribute of a style tag so that vue-jest can deal with external css.

https://github.com/vuejs/vue-jest/pull/158

PR #158, opened 2019-02-11
External css

hewIngram:
This addresses use of external css files for components as described here: https://github.com/vuejs/vue-jest/issues/76

@eddyerburgh not sure how you feel about just writing the external file into `stylePart` but it seemed like the easiest way of doing this.

### DIFF
--- a/e2e/__projects__/style/components/External.vue
+++ b/e2e/__projects__/style/components/External.vue
@@ -1,5 +1,13 @@
 <template>
-  <div :class="$style.testClass" />
+  <div :class="$style.testClass">
+    <div class="a" />
+  </div>
 </template>
 
 <style module src="./styles/external.css" />
+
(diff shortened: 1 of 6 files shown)


eddyerburgh:
Looks good :) 

It's fine to rewrite the content, that's how we do it for script and template.

One small refactor—can you use the [`loadSrc` function](https://github.com/vuejs/vue-jest/blob/master/lib/process.js#L31). You could move it into lib/utils to use it in both files.

hewIngram:
Ah I totally didn't spot that. 👍 
I also renamed filename to filepath as it is the path + more consistent with the rest of the code.

The `utils.js` file doesn't have unit tests, do you want me to write some or are you happy with the e2e coverage?

eddyerburgh:
The e2e tests are fine

PR #161, opened 2019-02-14
fix(ts-config): remove dependency on ts-jest and configure tsconfif w…

yohodopo:
…ith vue-jest

Remove dependency over ts-jest and let the user configure his tsconfig with `vue-jest` when specified.
When tsconfig/ts-jest is not configured then a will look for tsconfig at the root of the application.

Hey @eddyerburgh I've tried creating to use babel to compile typescript files here https://github.com/yohodopo/vuejest/blob/bcfa5bc337910f0ba86a93c42d660a5c2bfa74f7/lib/utils.js#L112-L131 but I'm not very confident if thats what we'd want to do that since users might want to use some of the features it does not offer yet with the list of plugins, It would also take away the most desired feature of typechecking.


eddyerburgh:
Thanks for this :)

The link you've posted is 404ing. Is it a private repo?

lib/utils.js
```diff
@@ -1,6 +1,8 @@
 const loadPartialConfig = require('@babel/core').loadPartialConfig
 const createTransformer = require('ts-jest').createTransformer
```
eddyerburgh: This will still error if users don't have TypeScript installed, can you move the require to inside the `getTSConfig` function

e2e/__projects__/tsconfig/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "babel-in-package",
```
eddyerburgh:
```suggestion
  "name": "tsconfig",
```

e2e/__projects__/tsconfig/test.js
```diff
@@ -0,0 +1,54 @@
+import { getTSConfig } from 'vue-jest/lib/utils'
```
eddyerburgh: This directory is intended for e2e tests. Can you move these tests into a unit test file: /lib/utils.spec.js, and create a `test:unit` script that runs files with a spec extension that aren't in /e2e

eddyerburgh:
Thanks for the work :)

I think it's fine to use Babel for TypeScript:

1. It's what Jest does now for TypeScript files by default
2. It removes tsconfig resolution logic from vue-jest
3. The features it doesn't support are minimal
4. The intention for vue-jest isn't to do type checking

I thought we could use babel in the beta version and get feedback from users. 

FYI The reason we started to use ts-jest originally was to move the logic for resolving tsconfig options to another package. But I'm not sure the package is going to be maintained, so we should remove it as a dependency. 

Can you open a new PR that uses babel for compiling TypeScript?

lmiller1990:
We fixed this in another PR.

PR #162, opened 2019-02-15
feat(babel-typescript) transpile typescript code using babel

yohodopo:


yohodopo:
@eddyerburgh , as discussed in #161 , I've made changes to use babel for transpiling typescript

lib/utils.js
```diff
@@ -114,12 +109,60 @@ const logResultErrors = result => {
...
 
+const getTSBabelOptions = () => {
+  const ts = getTsInstance()
+  const { compilerOptions } = getTsConfig()
+  const TSBabelOptions = {
+    presets: [
+      '@babel/env',
+      ['@babel/preset-typescript', { allExtensions: true }]
+    ],
+    plugins: [
```
eddyerburgh: Can you explain why we need each of these plugins?

lib/utils.js
```diff
@@ -114,12 +109,60 @@ const logResultErrors = result => {
...
+      ['@babel/plugin-proposal-decorators', { legacy: true }],
+      ['@babel/plugin-proposal-class-properties', { loose: true }],
+      '@babel/plugin-proposal-object-rest-spread',
+      '@babel/plugin-syntax-dynamic-import',
+      ['@babel/plugin-transform-runtime', { corejs: 2 }]
```
eddyerburgh: I can see why we're adding `@babel/plugin-proposal-decorators`, but why are we adding '@babel/plugin-proposal-object-rest-spread' and '@babel/plugin-transform-runtime'?

eddyerburgh:
Since babel-jest requires users to have babel-typescript in their babel options, I think we should too. 

In other words, can you make it so that ts code is compiled by babel. We could remove all the custom typescript-transformer code, and just use the babel-jest transformer to compile.

We'll need to tell users that they should add the correct plugins for compiling their TypeScript code.

leipert:
@eddyerburgh / @yohodopo I would be looking forward to help with this and get it into a new beta quickly. Is there anything I could do to help speeding this up?

package.json
```diff
@@ -67,12 +65,20 @@
...
+    "@babel/plugin-proposal-decorators": "^7.3.0",
+    "@babel/plugin-proposal-object-rest-spread": "^7.3.2",
+    "@babel/plugin-syntax-dynamic-import": "^7.2.0",
     "@babel/plugin-transform-modules-commonjs": "^7.2.0",
+    "@babel/plugin-transform-runtime": "^7.2.0",
+    "@babel/preset-env": "^7.3.1",
+    "@babel/preset-typescript": "^7.1.0",
+    "@babel/runtime-corejs2": "^7.3.1",
     "@vue/component-compiler-utils": "^2.4.0",
+    "babel-jest": "24.1.0",
```
leipert: @yohodopo / @eddyerburgh Is there a reason why we pin this to `24.1.0`? Should this maybe be a peer dependency and/or also point to a version range with `^`?

package.json
```diff
@@ -30,10 +30,8 @@
   "license": "MIT",
   "devDependencies": {
     "@babel/core": "^7.2.2",
-    "@babel/preset-env": "^7.2.3",
     "@vue/test-utils": "^1.0.0-beta.25",
     "babel-core": "^7.0.0-bridge.0",
```
leipert: @yohodopo / @eddyerburgh Also updating to `babel-jest@24` should enable us to remove the dependency on the babel-core bridge, as `babel-jest` supports `babel@7` now

yohodopo: I guess it's is because of our dependency on `ts-jest`. Now that ts-jest v24.0.0 is released, we no longer need this.

yohodopo:
@leipert , the biggest challenge I see here is that we cannot completely ignore typescript compilers as most of the engineers are still using `ts-jest` to transform their typescript code. So we might have to keep our code open for both the approaches. 

What I'm saying is we could add a check [here](https://github.com/vuejs/vue-jest/blob/master/lib/typescript-transformer.js) to check from the jest config if the users are using ts-jest or `babel-jest` to transform typescript files and do the same conditionally here.

Please let me know if you have any questions

lmiller1990:
I accomplished something similar here: https://github.com/vuejs/vue-jest/pull/239

Which also gives optional compile-time errors. Is there any benefit to going with babel?

PR #169, opened 2019-03-20
Upgrade ts-jest to 24.0.0

AlbertBrand:


icehaunter:
Could we maybe get a beta.3 with this fix, just to remove the error? :)

Issue #45, opened 2018-01-17
Failed parse using external template


mazipan:
I used external template like :
```html
<template
  src="vue2-simplert-core/simplert.html">
</template>
```
Success when building production file using `vue-loader@13.7.0`.
But unit test failed to run using `vue-jest@1.4.0`

Here some snippet of my Jest configuration : 
```javascript
 "jest": {
    "mapCoverage": true,
    "coverageDirectory": "test/coverage",
    "moduleNameMapper": {
      "^vue$": "vue/dist/vue.common.js"
    },
    "moduleFileExtensions": [
      "js",
      "vue"
    ],
    "transform": {
      "^.+\\.js$": "<rootDir>/node_modules/babel-jest",
      ".*\\.(vue)$": "<rootDir>/node_modules/vue-jest"
    }
  }
```

Error that I got is : 
```
ENOENT: no such file or directory, open '~/project/src/vue2-simplert-core/simplert.html'
```
if I look into error message, it because pointing to my project root folder instead of `node_modules` folder.

Is there any missing from my side ?

Thanks for all response 👍

eddyerburgh:
What's the absolute path of simplert.html?

mazipan:
It's coming from other npm repository, so should be coming from `node_modules` folder.

eddyerburgh:
Does this resolve correctly when you use vue-loader?

mazipan:
Sure, working like charm...
That's why I create this issue.
Maybe my bad, because I am not familiar with Jest configuration.
But I want to know from your side first. Maybe you have some suggestions.

eddyerburgh:
This is a problem on our side we need to fix. At the moment, we only look for a src that is relative, not in node_modules

Thanks for bring it to my attention, I'll look to get a fix in place 🙂

laggingreflex:
I'm using vue-loader but still can't seem to make a even relative external templates work

Repro: https://github.com/laggingreflex/repro-vue-jest-template

It gives the error

```
> jest


Cannot use <template> as component root element because it may contain multiple nodes.

 FAIL  src/test.js
  ● Test suite failed to run


    [vue-jest] Error: Vue template compilation failed

    > 1 | const App = require('./app').default;
        |                                        ^
      2 | 
      3 | describe('test', () => { it('test', () => {}) })
      4 | 

      at error (node_modules/vue-jest/lib/utils.js:102:9)
      at Object.<anonymous> (src/test.js:1:40)

```

[...]

PR #183, opened 2019-06-28
feat: Add external src resolver (#45)

robwozniak:


robwozniak:
related to issue #45

robwozniak:
I see CI issues, but I am wondering if these changes direction is a good approach to solve an issue. Could someone familiar with this repo advice?

PR #195, opened 2019-09-03
added support postcss

mrvasia:


mrvasia:
@eddyerburgh

lib/compilers/postcss-compiler.js
```diff
@@ -0,0 +1,28 @@
...
+  postcss(plugins)
+    .process(content, {
+      from: undefined
+    })
+    .then(result => {
+      css = result.css || ''
+    })
+    .catch((e) => {
+      css = ''
+      throwError(e)
```
eddyerburgh: Instead of throwing an error, can you log an error if the hideStyleWarn is false, like this: https://github.com/mrvasia/vue-jest/blob/dd57354402945c796466d33dc95d4d2d7f73babe/lib/compilers/sass-compiler.js#L29

lib/compilers/postcss-compiler.js
```diff
@@ -0,0 +1,28 @@
...
+const postcssrc = require('postcss-load-config')
+const ctx = { parser: true, map: 'inline' }
+const { plugins } = postcssrc.sync(ctx)
+
+const throwError = require('../throw-error')
+
+module.exports = (content) => {
+  let css = null
+
+  postcss(plugins)
```
eddyerburgh: You can run postcss synchronously: https://github.com/vuejs/component-compiler-utils/blob/37b4a6a99ecaa8bf27c777191ac17e9168c70cd1/lib/compileStyle.ts#L115

It means asynchronous plugins won't be applied, but I think that's is fine

lib/compilers/postcss-compiler.js
mrvasia: fixed

lib/compilers/postcss-compiler.js
mrvasia: I added check for async plugin, i think it is more correct way, because we have async plugin and processing postcss is crashed

mrvasia:
@eddyerburgh take a look again (:

mrvasia:
@eddyerburgh

mrvasia:
up @eddyerburgh

matthewmorek:
@eddyerburgh Any chance you could have a look at it, as it would be great to have proper support for PostCSS rather sooner than later. 😉

pushkarskiy:
Hello @eddyerburgh! What about merge pull request ? I see, mr is completed, or have some problem ? I think it is a good, solution, and  we await this feature in vue jest a lot of time.

lmiller1990:
Sorry this repo has been neglected for so long! Let's merge this. I don't know anything about PostCSS, but tests all seem to be passing.

I will update the readme noting we now have PostCSS support. Let's do a release in the next few days for 4.0.0-alpha.X.

lmiller1990:
It is out https://github.com/vuejs/vue-jest/releases/tag/v3.0.6

Issue #137, opened 2018-12-19
Sass Packages is Not Supported


evanleck:
Maybe there's a good reason for it, but it strikes me as odd that "node-sass" is required when [sass](https://www.npmjs.com/package/sass) is listed on the [installation page](https://sass-lang.com/install). I'm currently using sass with Parcel and installing both sass and node-sass isn't an option, so the requiring node-sass is making testing basically impossible for me. Would it be possible to use one or the other, or provide an option to skip style compilation altogether?

Thank you!

evanleck:
Well hot damn! Turning off CSS compilation did the trick, thank you!

Yeah, I realize node-sass is faster, I'm just trying to figure out why it would be required when sass should work just as well. The sass package [_is_ dart-sass](https://github.com/sass/dart-sass#from-npm) compiled to JavaScript so it technically is the "reference implementation of Sass". It seems like using the reference implementation should at least be an option when using this package.

Parcel has support for node-sass but [there's a bug](https://github.com/parcel-bundler/parcel/issues/1836) that's [mentioned in the docs](https://parceljs.org/scss.html) so they default to sass. Installing node-sass basically kicks Parcel over to use that instead of sass and there are oddities there that I'm trying to avoid by using sass.

Anyway, my issue was solved by disabling `experimentalCSSCompile` so thank you for pointing that out!

aldarund:
>TBH in my personal experience I haven'T come across a project using sass in ages

FYI -> vuetify 2 (currently in beta) require and requires specifically sass and not node-sass, so if someone want to use vuetify and sass in their project they are forced to also install node-sass just for vue-jest.

dkonchekov:
Hi!
vue-cli wizard allows to select the options: dart-sass, jest.
But this combination fails when CSS Modules are used because of node-sass requirement.
experimentalCSSCompile flag is not solution because we need test classes in jest.
Having both node-sass and sass discredits the idea of use sass instead node-sass.

You have @vue/component-compiler-utils@^2.4.0 dependency that in 3.x version has dropped node-sass in favor of sass. 

Have you any plans to update your @vue/component-compiler-utils?

I think it will be better if @vue/component-compiler-utils delegate choice of preprocessor implementation to user as it is done in sass-loader (https://github.com/webpack-contrib/sass-loader/blob/master/src/getDefaultSassImplementation.js) and move node-sass and sass to peerDependencies.

JessicaSachs:
[...]

PR #209, opened 2019-12-18
feat: Replace node-sass to sass

dyurkavets:
Replacing `node-sass` to `sass` since is the primary implementation and doesn't require to be compiled.

In addition I've updated `@vue/component-compiler-utils` to the latest `3.x` version that uses `sass` too


JessicaSachs:
Thank you so much @dyurkavets! Glad we could finally get around to merging this. Should close #137

PR #173, opened 2019-04-19
docs: add info for babel7 peer dep

escapedcat:
Every now and then I check if I still need the babel7 bridge and go through some issues of jest, babel-jest, etc. to figure why I still need it. I ended up here and saw #160.
It seems that a lot of people still being confused about this because a lot of people already moved to Babel v7.

It's fine that vue-jest needs the bridge, but might be easier for others to see immeditately what is needed to make it work. Seems like there's a bigger amount of people who don't look at the peer-deps first. Maybe this helps them.

PR #215, opened 2020-01-31
Update incorrect Babel 7 instruction in README.md

middric:
The readme doc incorrectly states that if you are using versions lower than 24.0.0 of Jest you should use the babel-core@bridge package. This should be versions _greater_ than 24.0.0

I spent a long time banging my head against the wall over this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ npm install --save-dev vue-jest
 
 ### Usage with Babel 7
 
-If you use [jest](https://github.com/facebook/jest) < 24.0.0 and [babel-jest](https://github.com/facebook/jest/tree/master/packages/babel-jest) make sure to install babel-core@bridge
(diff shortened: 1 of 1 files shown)


morficus:
man, this really needs to be merged. there goes an hour of my life I shall never get back.

why does this lib still require the use of `babel-core@bridge` and hasn't transitioned to `@babel/core`? I see `@bable/core` as a "peerDependency" so what's up wit that?

morficus:
@escapedcat @eddyerburgh  ☝️ 
(tagging you two since you were involved in updating the docs for Babel 7 compatibility)
any chance we could get this README issue merged to save others from future pain?

morficus:
same comment was left on the original PR as well: #173

escapedcat:
Sorry for only seeing this now but I can't do anything here :(

PR #220, opened 2020-03-07
fix: support dart-sass in addition to node-sass

JessicaSachs:
This should allow users to use dart-sass without requiring node-sass in their builds. The `@vue/component-compiler-utils` will figure out which one to use.

Related to #137


JessicaSachs:
Not going this route. Preferring dart-sass over node-sass because this is what compiler-utils does.

PR #227, opened 2020-03-10
feat: add support for vue-i18n custom blocks by default

JessicaSachs:
This PR adds support for an extensible API to parse and process custom blocks.

See the [rendered README](https://github.com/vuejs/vue-jest/pull/227/files?short_path=04c6e90#diff-04c6e90faac2675aa89e2176d2eec7d8) for API changes and writing a processor

```js
// jest.config.js
module.exports = {
  // ...
  globals: {
    'vue-jest': {
      transform: {
        'your-custom-block': './your-transformer.js'
      }
    }
  }
}
```


kazupon:
@JessicaSachs 
Thank you for your support !

I'm glad you support i18n custom blocks by default in vue-jest.
However, if we will consider the maintainability of vue-jest, I think that it's should provide an interface that supports transform to code for custom blocks.

vue-jest seem that have already a mechanism that user can transform to code.
You can see the code in `processStyle` of `lib/process-style.js`.

https://github.com/vuejs/vue-jest/blob/master/lib/process-style.js#L67-L93

In the above, the custom transformer path gets from the jest config and require it, and the code is generated by custom transformer.

As well as, I think we can do custom blocks.
The custom transformer need to provide with library author like vue-i18n, but I don't think it's a problem because it's the responsibility of side providing custom blocks.

JessicaSachs:
I agree! In discord I am asking for help to design an extensible interface
for custom blocks :-) come chat with me.

JessicaSachs:
OK! @kazupon, @lmiller1990, @afontcu, and @dobromir-hristov would love feedback from any and all of you 🙇‍♀

lib/transformers/i18n.js
```diff
@@ -0,0 +1,61 @@
...
+
+  if (block.attrs && block.attrs.src) {
+    content = fs.readFileSync(path.resolve(filename, '../', block.attrs.src))
+  }
+
+  if (langKnown) {
+    return JSON.stringify(JSON.parse(convert(content, langKnown)))
+  }
+
+  try {
```
lmiller1990: do we need a warning here? "Your file is not valid JSON or YAML".

lib/transformers/i18n.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+}
+
+module.exports = {
+  process(blocks, vueOptionsNamespace, filename) {
+    const base = `${vueOptionsNamespace}.${VUE_I18N_OPTION} = []`
+    const codes = blocks.map(block => {
+      if (block.type === 'i18n') {
+        const value = parseLanguageAndContent(block, filename)
+          .replace(/\u2028/g, '\\u2028')
```
lmiller1990: What do these codes represent?

lib/process-custom-blocks.js
```diff
@@ -0,0 +1,41 @@
+const { getVueJestConfig, getCustomTransformer } = require('./utils')
+const vueOptionsNamespace = require('./constants').vueOptionsNamespace
+
+function applyTransformer(
```
lmiller1990: Should we try to match the jest process api a bit more closer? That would be `applyTransform(blocks, filename, config, transformer, vueOptionsNamespace) I guess. Although maybe more awkard to call then - just thinking out loud

lib/generate-code.js
```diff
@@ -21,7 +24,7 @@ module.exports = function generateCode(
   }
 
   output +=
-    `var __options__ = typeof exports.default === 'function' ` +
+    `var ${namespace} = typeof exports.default === 'function' ` +
```
lmiller1990: Any specific reason to move the constant to another file? It seems to be used elsewhere when it wasn't before - why is that?

JessicaSachs: Because it was being hardcoded inside of vue-i18n-jest and we refactored this, which broke vue-i18n-jest if we were to have released v4 of vue-jest... We should make any required constants passed in rather than requiring developers to dig into vue-jest source code to figure out what constants we're using

lib/process-custom-blocks.js
JessicaSachs: I wasn't sure about this either. I think if there was no external context, like if jest process didn't exist, maybe an object-based api would be best, because there's a lot of optional params.

lib/transformers/i18n.js
JessicaSachs: They're line seperators, but really, I have no idea, I copied them from vue-i18n-jest

lib/transformers/i18n.js
JessicaSachs: Could do. I matched the existing code.

lib/process-custom-blocks.js
JessicaSachs: I wasn't sure about this either. I think if there was no external context, like if jest process didn't exist, maybe an object-based api would be best, because there's a lot of optional params.

lib/transformers/i18n.js
lmiller1990: ok, vue-i18n is battle tested, going to assume kazupon knows what he is doing

lib/process-custom-blocks.js
lmiller1990: Object based works well for me too - what do you think? I really like to use an object when there are more than 1-2 params, since it is easier to read.

If it's not too much trouble that change could go with that? I don't mind too much, though.

lib/generate-code.js
lmiller1990: Right, makes sense, sounds good

lib/process-custom-blocks.js
JessicaSachs: Updating it for this

kazupon:
@JessicaSachs 
Thanks a lot!.

I'm realy gload to support vue-i18n as default at vue-jest.
But the supporting at least one custom block will seem unfair to the creator of the library that provides the custom block.

I hope that vue-jest will offer I/F only for custome blocks without natively supporting vue-i18n. :)

JessicaSachs:
No problem. I’m happy to remove the processor natively now that we’re
supporting custom blocks generally.

JessicaSachs:
@kazupon Here is the [file](https://github.com/vuejs/vue-jest/blob/df2bfa241bb105b9a17d1529357e3a6d55ff1dad/lib/transformers/i18n.js
) tree with the final implementation of vue-i18n natively. As discussed, I removed it from this PR.

Please feel free to adapt this into your own package. I made some improvements on the existing wrapper you built, so please take a look 🙇‍♀

kazupon:
@JessicaSachs 
Okay! Thanks!

package.json
```diff
@@ -72,7 +73,8 @@
     "chalk": "^2.1.0",
     "convert-source-map": "^1.6.0",
     "extract-from-css": "^0.4.4",
-    "source-map": "^0.5.6",
```
sibiraj-sr: @JessicaSachs  this package is used here https://github.com/vuejs/vue-jest/blob/0b0d5e2dcd28da728bc0928efdfa6817df6113e7/lib/generate-source-map.js#L2

PR #231, opened 2020-03-23
chore: add dep

lmiller1990:
`source-map` was accidentally removed in here https://github.com/vuejs/vue-jest/commit/d6c4e475e5498083b6273d73e641a1e3f086d087#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L75

but we need it https://github.com/vuejs/vue-jest/blob/0b0d5e2dcd28da728bc0928efdfa6817df6113e7/lib/generate-source-map.js#L2

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "chalk": "^2.1.0",
     "convert-source-map": "^1.6.0",
     "extract-from-css": "^0.4.4",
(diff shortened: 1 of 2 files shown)


JessicaSachs:
Yikes! We need to get Circle CI back up and running.

PR #236, opened 2020-04-22
Support pure functional components

jesusgn90:
Before this PR, a component like the next one would fail to be tested because `$style` would be `undefined`:

```js
<script>
export default {
  name: 'FunctionalSFCRender',
  functional: true,
  render(createElement, { $style }) {
    return createElement('div', {
      class: [$style.ModuleClass]
    })
  }
}
</script>

<style lang="scss" module>
.ModuleClass {
  width: auto;
}
</style>
```

The reason for being a failure is related to a missing `<template>` block, however, a functional component can work properly using `functional: true` and a `render` function. Now it tries to find `functional: true` using a regex. 

This PR also includes a simple test that uses to fail without these changes.

### DIFF
--- a/e2e/__projects__/basic/components/FunctionalSFCRender.vue
+++ b/e2e/__projects__/basic/components/FunctionalSFCRender.vue
@@ -1,0 +1,17 @@
+<script>
+export default {
+  name: 'FunctionalSFCRender',
+  functional: true,
+  render(createElement, { $style }) {
+    return createElement('div', {
+      class: [$style.ModuleClass]
+    })
+  }
+}
+</script>
+
+<style lang="scss" module>
+.ModuleClass {
+  width: auto;
+}
+</style>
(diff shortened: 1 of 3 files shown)


lmiller1990:
Hi! Can we get a test for this?

For CI, try installing `source-map` at `0.5.6`. The newer one uses a promise which breaks things. This is working for me but I did not push this fix yet.

jesusgn90:
> Hi! Can we get a test for this?
> 
> For CI, try installing `source-map` at `0.5.6`. The newer one uses a promise which breaks things. This is working for me but I did not push this fix yet.

I've added a test at https://github.com/vuejs/vue-jest/pull/236/commits/9d28943e78dd60af2d7edf171bf38f5c9aaaa9ed, however, I couldn't test it locally, seems like a pain to make it run properly so I don't know if the test is working (it should though). Also, the problem with the CI which I don't know about it so let me know if you need anything else from my side, regards.

lmiller1990:
Thanks @jesusgn90 , I will review and get this merged this weekend - been pretty busy. I can fix the CI for you too, no problem.

PR #239, opened 2020-04-25
feat: type check vue files

lmiller1990:
Not super clean - creates a tmp file (but the OS will normally clear these periodically. Should confirm this isn't going to be a problem).

At the very least a decent POC. Hidden behind an `experimentalTs` flag. Makes tests run a bit slower on basic observation. Eg

```
vue-jest: {
  enableExperimentalTsDiagnostics: true
}

```


lib/transformers/typescript.js
```diff
@@ -16,6 +52,21 @@ module.exports = {
...
 
+    if (vueJestConfig.enableExperimentalTsDiagnostics) {
+      const diagnostics = getDiagnostics(
+        scriptContent,
+        { ...tsconfig, noEmitOnError: true },
+        typescript
+      )
+      logDiagnostics(filePath, diagnostics, typescript)
+
+      if (diagnostics.length) {
```
ahnpnl, line 63: You can put warning message here, that is how currently ts-jest [does](https://github.com/kulshekhar/ts-jest/blob/452639242d7f59502bda7aa94b81735f9c9ad203/src/config/config-set.ts#L382) with `isolatedModules: true`

lib/transformers/typescript.js
```diff
@@ -5,9 +5,45 @@ const {
...
-  getVueJestConfig
+  getVueJestConfig,
+  warn
 } = require('../utils')
 
+function getDiagnostics(tsBlob, options, ts) {
+  ensureRequire('tempy', ['tempy'])
+  const tempy = require('tempy')
+
+  const temp = tempy.writeSync(tsBlob, { extension: 'ts' })
```
ahnpnl, line 16: If you use in-memory approach, you won’t need this

lib/transformers/typescript.js
```diff
@@ -16,6 +52,21 @@ module.exports = {
     const tsconfig = getTsJestConfig(config)
     const babelOptions = getBabelOptions(filePath)
 
+    if (vueJestConfig.enableExperimentalTsDiagnostics) {
+      const diagnostics = getDiagnostics(
+        scriptContent,
+        { ...tsconfig, noEmitOnError: true },
```
ahnpnl, line 58: `ts-jest` config always:

-  set `sourceMap` to [true](https://github.com/kulshekhar/ts-jest/blob/452639242d7f59502bda7aa94b81735f9c9ad203/src/config/config-set.ts#L590). Since you don't need `sourceMap` here, you can try setting it to `false` to TypeScript `Program` won't try to emit source map -> improve performance. 

- set `noEmit` to [false](https://github.com/kulshekhar/ts-jest/blob/452639242d7f59502bda7aa94b81735f9c9ad203/src/config/config-set.ts#L595). Probably here you can try setting it to `true` so `program.emit()` won't try to emit any files. -> improve performance.

When using `program.emit()`, TypeScript will also try to emit files, which can include compiled js, source map and type declaration.

lib/transformers/typescript.js
```diff
@@ -5,9 +5,45 @@ const {
...
+function getDiagnostics(tsBlob, options, ts) {
+  ensureRequire('tempy', ['tempy'])
+  const tempy = require('tempy')
+
+  const temp = tempy.writeSync(tsBlob, { extension: 'ts' })
+  let program = ts.createProgram([temp], options)
+  program.emit()
+
+  let allDiagnostics = ts
+    .getPreEmitDiagnostics(program)
```
ahnpnl, line 21: `getPreEmitDiagnostics` is usually slow. Probably you can try to use [this](https://github.com/kulshekhar/ts-jest/blob/452639242d7f59502bda7aa94b81735f9c9ad203/src/compiler/language-service.ts#L21)

lmiller1990:
@ahnpnl sure, it would be great to use `ts-jest`. What are you thinking?

If you are interested you can most certainly make a PR targeting this branch. In the meantime, I will make the changes you suggested (sometime in the next day or two, been busy...)

ahnpnl:
> @ahnpnl sure, it would be great to use `ts-jest`. What are you thinking?
> 
> If you are interested you can most certainly make a PR targeting this branch. In the meantime, I will make the changes you suggested (sometime in the next day or two, been busy...)

I will explain my idea first. 

Currently I see in frontend world, we have UI libraries/frameworks which follow traditional file extensions like `js`, `jsx`, `ts`, `tsx` **vs** UI libraries/frameworks which follow custom file extensions like `vue`, `svelte`. Normally `ts-jest` takes care of transforming and type checking with traditional file extensions only. I want to make a new feature that `ts-jest` can support transforming and type checking with custom file extensions.

So how it affects to UI libraries/frameworks like Vue, Svelte.. when it comes to Jest transformer and TypeScript support. The idea is SoC concept, in which transformers like `vue-jest` does what it does best and `ts-jest` does what it does best:

- `vue-jest` takes care of transforming all other things rather than what is inside `<script>`.

- `ts-jest` takes care of transforming only what is inside `<script>`.

**Proposed implementation**

- `vue-jest` makes a chain of transformers invoking, you can reference `ts-jest` [here](https://github.com/kulshekhar/ts-jest/blob/8d46febf246aa1a00980851ef376c4ff7223380f/src/ts-jest-transformer.ts#L120)

- `vue-jest` will deliver the content of `<script>` to `ts-jest` and let `ts-jest` handling transforming + type checking.

- After `ts-jest` finishes, it will deliver back to `vue-jest` the output as well as throwing error for type checking (`isolatedModules: false`) or show warning for type checking (`isolatedModules: true`) and `vue-jest` can do the rest of the work.

lmiller1990:
This is a good idea @ahnpnl. I am going to play around in a new repo: https://github.com/lmiller1990/vue-ts-jest

There is also something new we need to support in Vue jest: https://github.com/vuejs/rfcs/blob/sfc-improvements/active-rfcs/0000-sfc-script-setup.md#transform-api

Is there a good place I can message you in real time (discord, etc)?

ahnpnl:
Nice, I like that TransformAPI.

Feel free to ping me on Discord at Ahn#1702

lmiller1990:
We will continue this work in https://github.com/lmiller1990/vue-ts-jest

PR #250, opened 2020-07-02
Force ES5 for TS compilation

lmiller1990:
This prevents `const vue_1 = require('vue')` from conflicting, since it will do `var vue_1` instead.

This feels pretty bad, I don't know the correct solution. 🤔

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -67,7 +67,8 @@ const getTsJestConfig = function getTsJestConfig(config) {
   const createTransformer = require('ts-jest').createTransformer
   const tr = createTransformer()
   const { typescript } = tr.configsFor(config)
-  return { compilerOptions: typescript.options }
+  // Force es5 to prevent const vue_1 = require('vue') from conflicting
+  return { compilerOptions: { ...typescript.options, target: 'es5' } }
 }
 
 function isValidTransformer(transformer) {


LinusBorg:
I think we could use a pretty basic regex to change that line to a `var ...` somewhere around here:

https://github.com/vuejs/vue-jest/blob/e0cbca7920e3afca3bd333377a4e463b57631e66/lib/generate-code.js#L22-L25

LinusBorg:
This whole part is a bit brittle since we use code that is meant to end up in multiple (virtual)  files and concat it into one piece of code.

If I were to have a top-level variable or function called `render` in my script block, the `function render(` generated by the template compilation step would conflict here as well, I would assume.

lmiller1990:
I would say so - I experimented a bit to  try and get other variables to conflict, no luck, but I'm sure it's possible.

I pinged a maintainer of ts-jest to get some ideas, he may have some experience around this. I am not really sure on the best way to solve this problem 🤔  what we do atm definitely feels pretty dirty.

ahnpnl:
Hi, come back to this I have a suspicion. It seems like the error is thrown because in the final `js` output which is given to `jest` containing twice `const vue_1 = require(‘vue’)` which will then cause the issue cannot redeclare block-scope variable.

If that is the case, the `const vue_1 = require(‘vue’)` needs to be removed from `vue-jest` transpile output by either of these ways:

- Use regex
- Write a Babel plug-in, which is also an AST transformer to use with Babel.
- Write a TypeScript transformer, which is also AST transformer to use with TypeScript API.

I’d recommend to use AST transformer over regex when dealing with transpiring.

lmiller1990:
I think you are right - using `var` is just a work around, not really a good solution. 

I think that we should make a new codebase for a TypeScript and ts-jest based solution that implements the best patterns that I learned from looking at other jest transformers, and pin it to Jest versions like ts-jest does - supports both Vue 2.x and 3.x, and multiple jest versions, is very difficult in a single codebase.

ahnpnl:
ya, agree. Maybe you can setup the base structure 1st and others can start contributing. Maybe worth to write down in an issue about all the plans :)

lmiller1990:
Yep, we definitely need a good "plan of attack".

PR #259, opened 2020-07-29
Fix bare imports of Sass files with dots.in.filename.scss

vvanpo:
As discovered in https://github.com/vuejs/vue-jest/pull/256#issuecomment-664764912.

When importing a Sass file that has dots in its filename but without the `.scss` extension (e.g. `@import 'foo.bar';`), our resolver would assume `.bar` was the extension and forego appending `.scss`, causing `require.resolve()` to throw an exception.

We need to look for `.scss`, `.sass`, and `.css` extensions, as all of these are possible. Previously, it would automatically assume the extension for the file was the same as the `lang` attribute set on the `<style>` block, but Sass has no problems importing `.sass` from a `.scss` file or vice-versa, and `.css` is also very common.

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,11 +298,23 @@ If a string is provided, it will be an assumed path to a TypeScript configuratio
 ### Supported style languages
 
 - **stylus** (`lang="stylus"`, `lang="styl"`)
-- **sass** (`lang="sass"`)
-  - The SASS compiler supports jest's [moduleNameMapper](https://facebook.github.io/jest/docs/en/configuration.html#modulenamemapper-object-string-string) which is the suggested way of dealing with Webpack aliases.
+- **sass** (`lang="sass"`), and
 - **scss** (`lang="scss"`)
 
-  - The SCSS compiler supports jest's [moduleNameMapper](https://facebook.github.io/jest/docs/en/configuration.html#modulenamemapper-object-string-string) which is the suggested way of dealing with Webpack aliases.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -298,11 +298,23 @@ If a string is provided, it will be an assumed path to a TypeScript configuratio
...
 - **scss** (`lang="scss"`)
 
-  - The SCSS compiler supports jest's [moduleNameMapper](https://facebook.github.io/jest/docs/en/configuration.html#modulenamemapper-object-string-string) which is the suggested way of dealing with Webpack aliases.
+  - The Sass compiler supports Jest's [moduleNameMapper](https://facebook.github.io/jest/docs/en/configuration.html#modulenamemapper-object-string-string) which is the suggested way of dealing with Webpack aliases. Webpack's `sass-loader` uses a [special syntax](https://github.com/webpack-contrib/sass-loader/blob/v9.0.2/README.md#resolving-import-at-rules) for indicating non-relative imports, so you'll likely need to copy this syntax into your `moduleNameMapper` entries if you make use of it. For aliases of bare imports (imports that require node module resolution), the aliased value must also be prepended with this `~` or `vue-jest`'s custom resolver won't recognize it.
+    ```json
+    {
+      "jest": {
+        "moduleNameMapper": {
+          "^~foo/(.*)": "<rootDir>/foo/$1",
+          // @import '~foo'; -> @import 'path/to/project/foo';
```
vvanpo, line 310: I see that GitHub does not like my usage of comments in a JSON block, fair enough, I'll have to change this to a `js` block.

lmiller1990: Yep, comments are not valid json. That's why!

lmiller1990:
I will release both a new version for 4.x and 5.x this weekend.

lmiller1990:
Latest 4.x and 5.x is out.

PR #262, opened 2020-07-30
refactor!: support wider range of jest versions by making babel-jest & ts-jest peer dependencies

sodatea:
It is mainly because ts-jest has a hard-coded jest version requirement.

And ts-jest is listed as optional because not every project needs it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "@vue/compiler-sfc": "3.0.0-alpha.10",
     "@vue/test-utils": "^1.0.0-beta.25",
     "babel-core": "^7.0.0-bridge.0",
-    "babel-jest": "^23.6.0",
+    "babel-jest": "^24.9.0",
     "coffeescript": "^2.3.2",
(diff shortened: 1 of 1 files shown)


sodatea:
@lmiller1990 Could this PR be backported to v4? Because ts-jest has a hard peer dependency on typescript, therefore may cause trouble for non-TS users.

lmiller1990:
Sure, I can do it.

PS: can we bump vue-cli to use v4 by default? v3 is not really worked on anymore, I think it's still the default, though.

lmiller1990:
I will track in here: https://github.com/vuejs/vue-jest/issues/281

sodatea:
Yeah, I plan to bump it in [the `next` branch](https://github.com/vuejs/vue-cli/tree/next).

lmiller1990:
Just to confirm - I was talking about making the default for Vue 2 be 4.x. Vue 3 will use 5.x

We don't support `<script setup>` yet - but I have a POC that will hopefully be a PR soon.

PR #263, opened 2020-08-02
Add coverage tool

vvanpo:
See the coverage report by running `yarn test:e2e && yarn coverage`. If you're on macOS, your browser should open automatically with the report. If you're on Linux, you'll have to open the `coverage/index.html` manually.

I was having a hard time picturing just what our test coverage was through our custom testrunner, so I set out to add support for automated coverage collection.

Because the testrunner runs a separate Jest process inside each test subpackage, and because Jest cannot be configured to collect coverage outside the package it is running in, I had to add `nyc` (Istanbul's CLI) and proxy the call to Jest through it. Then I needed to use Istanbul's API to merge the resulting coverage reports into a single summary.

I removed most of the dependencies (including Jest itself) from the test subpackages, because there is not much reason to duplicate these as node module resolution has no problem finding them in the root package, and the test suites are an order of magnitude faster this way. I'm tempted to just remove all dependencies from each subpackage and just install them on the root instead, then we could eliminate the `yarn install` step altogether if we wanted. If we for some reason need to test specific dependency versions differently between test suites, we can always add dependencies back in, and this also helps us be very explicit in that circumstance.

The only difference in test behaviour now that we're resolving most dependencies from the root `node_modules` is that these dependencies are now locked in the lockfile (they were before too, but that was because we were erroneously removing `yarn-lock.json` instead of `yarn.lock`; I've fixed that in this PR). This helps us have more deterministic builds, but if we don't actually want that and want to always test the latest version of each package it's simple enough to remove the lockfile and run `yarn install --no-lockfile` in CI.

Screen Shot 2020-08-01 at 7 56 34 PM


lmiller1990:
Wow, awesome! Sorry I just saw this. Please give me few days to review. I am not sure if there is a specific reason we have multiple projects with their own `package.json` files - I just joined this repo relatively recently, and am just trying to get everything back on track. I like these changes and anything that makes the repo more reliable and maintainable.

lmiller1990:
I am sorry - I completely forgot about this. I will try to review asap.

lmiller1990 (approved):

lmiller1990:
This looks fine. Can you rebase? Sorry it took me a long time to get to this! I really appreciate your contribution.

If you are excited about Vue jest I could really use some help on 5.0.0 for Vue 3.

nogic1008:
This PR makes e2e project's coverage report, not vue-jest itself.
We should make sure that the coverage is calculated correctly, but I don't think you need to make a report.

This PR is staled and conflicts with HEAD.
If you find it beneficial to create a report, please create another PR.

PR #264, opened 2020-08-05
Use transpileModule for TS and transform for babel conditionally

lmiller1990:
Seems to fix https://github.com/vuejs/vue-cli/issues/5714. Tested locally with latest vue-cli with a TS + babel and no babel setup.

### DIFF
--- a/lib/process.js
+++ b/lib/process.js
@@ -2,6 +2,7 @@ const { parse, compileTemplate } = require('@vue/compiler-sfc')
 const { transform } = require('@babel/core')
 const convertSourceMap = require('convert-source-map')
 const babelTransformer = require('babel-jest')
+const { transpileModule } = require('typescript')
 
 const generateSourceMap = require('./generate-source-map')
 const typescriptTransformer = require('./transformers/typescript')
@@ -9,6 +10,7 @@ const coffeescriptTransformer = require('./transformers/coffee')
 const _processStyle = require('./process-style')
 const processCustomBlocks = require('./process-custom-blocks')
 const getVueJestConfig = require('./utils').getVueJestConfig
+const getTsJestConfig = require('./utils').getTsJestConfig
 const logResultErrors = require('./utils').logResultErrors
 const stripInlineSourceMap = require('./utils').stripInlineSourceMap
 const getCustomTransformer = require('./utils').getCustomTransformer
@@ -68,10 +70,19 @@ function processTemplate(template, filename, config) {
 
   logResultErrors(result)
 
-  const babelify = transform(result.code, { filename: 'file.js' })
+  const tsconfig = getTsJestConfig(config)
 
-  return {
-    code: babelify.code
+  if (tsconfig) {
+    // they are using TypeScript.
+    const { outputText } = transpileModule(result.code, { tsconfig })
+    return { code: outputText }
+  } else {
+    // babel
+    const babelify = transform(result.code, { filename: 'file.js' })
+
+    return {
+      code: babelify.code
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


lib/process.js
```diff
@@ -68,10 +70,19 @@ function processTemplate(template, filename, config) {
...
 
-  const babelify = transform(result.code, { filename: 'file.js' })
+  const tsconfig = getTsJestConfig(config)
 
-  return {
-    code: babelify.code
+  if (tsconfig) {
+    // they are using TypeScript.
+    const { outputText } = transpileModule(result.code, { tsconfig })
+    return { code: outputText }
```
ahnpnl, line 78: If it is a combination of ts and Babel, I think you should do transform twice, once with typescript and once with Babel. Should it be like that ?

ahnpnl:
One of the differences between transforming with ts vs Babel is: ts doesn’t include polyfills while Babel does.

An example is: Array.prototype.flat which is available on node 11. When transforming ts to js, ts won’t polyfill it. If chaining Babel into the transform pipe, Babel will polyfill it.

lib/process.js
lmiller1990: Hm, I am not sure in this particular case it makes any difference. Not we are transforming the code generated by the `@vue/sfc-compiler`, not an actual human, so I am not sure we need to care about polyfills. I could be wrong.

lmiller1990:
Hm, interesting. I am not sure we need the double transform here; this library (vue-jest) doesn't have support for running in a browser right now, so everyone is running in a node + jsdom environment. Unless you are on a fairly old node version, the code generated from `@vue/compiler-sfc`  should work just fine without polyfills.

We could run it through babel to be sure, I guess - I wonder what kind of performance hit this would incur, though? For some very large test suites, this might not be ideal.

ahnpnl:
I think it’s safe to have e2e test for different combinations

lmiller1990:
Looks like our e2e tests are failing!

sodatea:
Have you tried optional chaining syntax in the template with Vue CLI TS + Babel setup?

sodatea:
(with Node.js v12)

lmiller1990:
I have not tried that, is there any specific thing I should be trying to chain? I have actually never used the optional chaining feature in Vue CLI, what chaining would you expect to potentially break things?

sodatea:
IIRC, in that setup, TS preserves the `?.` syntax, and Babel will then transpile it. So skipping Babel transpilation may break it in older Node versions. (Node 14 supports this syntax natively so we need to test in Node 12)

lmiller1990:
Oh I thought you were referring to how you can chain the webpack configuration in `vue.config.js`. I understand now. I will test that.

lmiller1990:
@sodatea I tried both with the null operator and (somewhat surprisingly) it worked fine.

I tried:
- [x] ts + babel
- [x] ts only
- [x] babel only

Here is the test I used:

```vue
<template>
  <div class="hello">
    <h1>{{ msg?.english }}</h1>
  </div>
</template>

<script lang="ts">
import { defineComponent } from 'vue';

export default defineComponent({
  name: 'HelloWorld',
  props: {
    msg: Object
  },
});
</script>
```

```ts
import { shallowMount } from '@vue/test-utils'
import HelloWorld from '@/components/HelloWorld.vue'

describe('HelloWorld.vue', () => {
  it('renders props.msg when passed', () => {
    const english = 'new message'
    const wrapper = shallowMount(HelloWorld, {
      props: { msg: { english } }
    })
    expect(wrapper.text()).toMatch(english)
  })
})
```

I am surprised babel knew about the null operator. 

Can you think of anything else I should test out?

sodatea:
Good to know that!
I think it's good to go now.

PR #273, opened 2020-09-01
docs: fix custom-block processor sample

nogic1008:
`process()` sample in README is different from the [implementation](https://github.com/vuejs/vue-jest/blob/master/lib/process-custom-blocks.js#L11).
This PR fixes it.


lmiller1990:
thanks!

PR #286, opened 2020-09-24
fix: external css files src resolution

FRSgit:
Follow up of #158 PR - external css files might use paths which needs `moduleNameMapper` resolution.

In case of the app I'm working on, we use external SCSS in a following way:
```html
<style lang="scss" src="@/assets/stylesheets/_bootstrap.module.scss" module="$styleBootstrap" />
```

`vue-jest` have failed in this case, because it always tried to load `style.src` as a relative path.
Now, the `moduleNameMapper` is used to resolve correct, absolute path before anything else happens.

Change have been tested - I think this one, simple test is enough to cover it, what do you think?

### DIFF
--- a/e2e/__projects__/style/components/External.vue
+++ b/e2e/__projects__/style/components/External.vue
@@ -6,6 +6,8 @@
 
 <style module src="./styles/external.css" />
 
+<style module="$style2" src="~__styles/external.css" />
+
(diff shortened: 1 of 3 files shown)


lmiller1990:
Simple enough, seems low risk. thanks!

PR #287, opened 2020-09-24
fix: external css properly handle html5-compliant syntax

FRSgit:
`vue-jest` fails when using proper html5 syntax when including external css files.

When I import external CSS in a following way (that's something recommended by [vue-loader](https://vue-loader.vuejs.org/spec.html#src-imports) and only way handled properly by vetur - [they do not support highlighting of self-closing `style` tags](https://github.com/vuejs/vetur/issues/1792)):
```html
<style src="./_bootstrap.module.scss" module="$styleBootstrap"></style>
```

`vue-jest` fails - it treats this style block as there would be no `src` attribute at all!

Fixed that case (added `content.trim()` - that's because `vue/component-compiler` for `<style></style>` returns bunch of spaces - something like: "          ").

### DIFF
--- a/e2e/__projects__/style/components/External.vue
+++ b/e2e/__projects__/style/components/External.vue
@@ -8,6 +8,8 @@
 
 <style module="$style2" src="~__styles/external.css" />
 
+<style module="$style3" src="./styles/external.css"></style>
+
 <style module="css">
 .a {
   background: color(red a(90%));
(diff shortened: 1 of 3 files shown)


lmiller1990:
Great, will need rebase (conflicts).

FRSgit:
Yea, sure, didn't want to ship these two things (this one & #286) in a single PR, so the conflicts were inevitable here.
It's 2am for me now, will do this tomorrow 😄

lmiller1990:
No rush, I tend to releases on fri/sat anyway, so I can test it out a bit!

FRSgit:
@lmiller1990 done!

lmiller1990:
I will release early next week.

PR #288, opened 2020-09-28
feat: add compile template options

### LINKED ISSUE #98: Tagged template strings are not supported
If tagged template strings are used, test suit will fail with following error:
```
Tagged template strings are not supported. Use `transforms: { templateString: false }` to skip transformation and disable this error, or `transforms: { dangerousTaggedTemplateString: true }` if you know what you're doing (1:303)

      at Node.initialise (node_modules/vue-template-es2015-compiler/buble.js:12998:10)
          at Array.forEach (<anonymous>)
          at Array.forEach (<anonymous>)

```
Apparently, this error message is from `buble.js`.
It seems that `vue-jest` uses `vue-template-es2015-compiler`, and  `vue-template-es2015-compiler` uses `buble`.
But neither `vue-jest` nor `vue-template-es2015-compiler` explains whether it is necessary and how to configure `buble`.

Is it possible to resolve this in `vue-jest`?
Any fixes or explanations are appreciated. Thank you.

I created a sample project to illustrate the issue. Steps to reproduce:
1. clone https://github.com/mjss/vue-jest-with-tagged-template
2. run `npm test:unit`

FRSgit:
Fixes #98.

### What?

This PR is adding to the `vue-jest` possibility to configure the template compiler used by [`@vue/component-compiler-utils` package](https://www.npmjs.com/package/@vue/component-compiler-utils).

### Why?

By passing correct configuration to the template compiler we can mitigate issues like the #98 - where the problem is directly connected with the proper tool configuration (in this case - the [`buble` library](https://www.npmjs.com/package/buble)).
BTW. the graphql case was added in this PR as an e2e test.

### Reproduction scenario

The problem can be reproduced by checking out the first commit in this PR and running `yarn test`. Then you'll see this error in the console:
```bash
FAIL ./test.js
  ● Test suite failed to run

    Tagged template strings are not supported. Use `transforms: { templateString: false }` to skip transformation and disable this error, or `transforms: { dangerousTaggedTemplateString: true }` if you know what you're doing (1:86)

      at Node.initialise (../../../node_modules/vue-template-es2015-compiler/buble.js:16016:10)
          at Array.forEach (<anonymous>)
          at Array.forEach (<anonymous>)
          at Array.forEach (<anonymous>)
          at Array.forEach (<anonymous>)
```

### DIFF
--- a/e2e/__projects__/basic/components/TemplateString.vue
+++ b/e2e/__projects__/basic/components/TemplateString.vue
@@ -1,0 +1,21 @@
+<template>
+  <div
+    :data-sth="
+      gql`
+        query {
+          id
+        }
+      `
+    "
+  />
+</template>
+
+<script lang="ts">
+export default {
+  methods: {
+    gql([str]) {
+      return str
+    }
+  }
+}
+</script>
--- a/e2e/__projects__/basic/package.json
+++ b/e2e/__projects__/basic/package.json
@@ -38,6 +38,13 @@
       "vue-jest": {
         "pug": {
           "basedir": "./"
+        },
+        "templateCompiler": {
+          "transpileOptions": {
+            "transforms": {
+              "dangerousTaggedTemplateString": true
+            }
+          }
         }
       }
     }
--- a/e2e/__projects__/basic/test.js
+++ b/e2e/__projects__/basic/test.js
@@ -1,5 +1,6 @@
 import { mount } from '@vue/test-utils'
 import TypeScript from './components/TypeScript.vue'
+import TemplateString from './components/TemplateString.vue'
 import { resolve } from 'path'
 import { readFileSync } from 'fs'
 import jestVue from 'vue-jest'
@@ -79,6 +80,15 @@ test('processes .vue file with lang set to coffeescript', () => {
 test('processes .vue files with lang set to typescript', () => {
   const wrapper = mount(TypeScript)
   expect(wrapper.element.tagName).toBe('DIV')
+})
+
+test('processes .vue files with template strings in the template', () => {
+  const wrapper = mount(TemplateString)
+  expect(wrapper.attributes('data-sth')).toBe(`
+      query {
+        id
+      }
+    `)
 })
 
 test('processes functional components', () => {
--- a/lib/process.js
+++ b/lib/process.js
@@ -57,16 +57,21 @@ function processTemplate(template, filename, config) {
     template.content = loadSrc(template.src, filename)
   }
 
+  const userTemplateCompilerOptions = vueJestConfig.templateCompiler || {}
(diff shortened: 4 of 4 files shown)


lmiller1990:
Great, sounds good.

We can release this week.

I wonder if we need something similar for Vue 3 (next branch...)

FRSgit:
@lmiller1990 No idea unfortunately - didn't have time to get a grasp of Vue 3 yet (other than reading through the breaking changes/migration guides 😄 ).

It depends if the [`buble`](https://www.npmjs.com/package/buble) is still being used anywhere in the package dependencies (on this branch the `@vue/component-compiler-utils` is importing `vue-template-es2015-compiler` which uses `buble`).
If there would be any way of getting rid of `buble` at all (I think that now only `vue-template-es2015-compiler` still uses it) then the configuration option added by this PR wouldn't be needed at all (babel supports configuration via `.babelrc`) :<

brunomperes:
Thanks a lot for this! Was this included in any stable release? I could not find it in the [releases](https://github.com/vuejs/vue-jest/releases) tab

FRSgit:
@brunomperes Unfortunately not yet - in my project I've added `vue-jest` dependency directly via github link `https://github.com/vuejs/vue-jest`.

@lmiller1990 What do you think about releasing new version?

lmiller1990:
Oops, sorry! Here we go: https://github.com/vuejs/vue-jest/releases/tag/v4.0.0-rc.1

FRSgit:
Thanks!
Hard to say when there is no roadmap, are there any other missing features/bigger bugs (other than the `ts-jest` thing which already has a PR)?
I would probably go through the PRs/issues, triage them, add the important ones (especially ones which would need making changes to the library's API) to the 4.0.0 milestone and just release it when they're resolved. I don't find `4.0.0` unstable at this point, so the other way would be to just release it now and continue with minor/patch release cycle every time a PR/bunch of PRs gets merged to the master.
I could give you a hand with this process, but probably you would make it much faster since you know this repo much better than me 🤷

lmiller1990:
I will try to get the`ts-jest` thing merged soon. Then do another release - that could be the final rc. Then 4.0.0 stable.

Triaging the issues is quite hard, since we have both v4 and v5 here, it's quite confusing. I am not sure how much value there is in triaging old issues - the time is better spent on other things, like merging getting 4.0 and 5.0 stable. 

For now, I will focus on getting the outstanding PRs merged. Thanks for your help!

FRSgit:
Makes sense. But well, at some point going through left-over issues is a must have - in the best case just to close them 😄 There are lots of issues on this repo and some of them are probably already outdated ;/
Nevertheless, feel free to ping me anywhere you see I could be of help!
Cheers!

PR #299, opened 2020-11-14
Add support for class components

### LINKED ISSUE #297: Class component appears not working with Vue 3.
Class component appears not working with Vue 3.

Vue CLI basic unit test is broken out of the box.

Repro: new project w/ Vue CLI, class component + ts + testing.

Stijn98s:
Closes #297

### DIFF
--- a/lib/generate-code.js
+++ b/lib/generate-code.js
@@ -28,8 +28,6 @@ module.exports = function generateCode(
 
   var tempOutput = node.toString()
 
-  console.log(tempOutput)
-
   if (
     tempOutput.match(/\}\(.*.?Vue\);/) &&
     tempOutput.includes('vue-class-component')
--- a/lib/generate-code.js
+++ b/lib/generate-code.js
@@ -28,10 +28,20 @@ module.exports = function generateCode(
 
   var tempOutput = node.toString()
 
+  console.log(tempOutput)
+
+  if (
+    tempOutput.match(/\}\(.*.?Vue\);/) &&
+    tempOutput.includes('vue-class-component')
+  ) {
+    node.add(';exports.default = {...exports.default.__vccBase};')
+  }
+
   if (tempOutput.includes('exports.render = render;')) {
     node.add(';exports.default = {...exports.default, render};')
   } else {
     node.add(';exports.default = {...exports.default};')
   }
+
   return node.toStringWithSourceMap({ file: filename })
 }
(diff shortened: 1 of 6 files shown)


Stijn98s:
@lmiller1990 what do you think about this fix?

lib/generate-code.js
```diff
@@ -28,10 +28,20 @@ module.exports = function generateCode(
 
   var tempOutput = node.toString()
 
+  console.log(tempOutput)
+
+  if (
+    tempOutput.match(/\}\(.*.?Vue\);/) &&
+    tempOutput.includes('vue-class-component')
+  ) {
+    node.add(';exports.default = {...exports.default.__vccBase};')
```
lmiller1990, line 37: I think we will now end up with 2x `node.add(';exports.default = {...exports.default};')` (added here *and* line 43).

Can you 1) check if that is the case and 2) make sure we only add it once? If we aren't careful, this can become really complex really fast!

lmiller1990 (commented):
Nice job, logic seems fine (bit dirty, but 🤷‍♂️ I don't see a cleaner way right now).

Can you address the one comment I left? Thanks!

lib/generate-code.js
Stijn98s: node now ends with
```
';exports.default = {...exports.default.__vccBase};',
';exports.default = {...exports.default, render};' 
```
or 

```
';exports.default = {...exports.default.__vccBase};',
';exports.default = {...exports.default};' 
```

I think this is the desired behavoir because otherwise the code will be more complex.

Example:
```
if (
  tempOutput.match(/\}\(.*.?Vue\);/) &&
  tempOutput.includes('vue-class-component')
) {
if (tempOutput.includes('exports.render = render;')) {
    node.add(';exports.default = {...exports.default.__vccBase, render};')
  } else {
    node.add(';exports.default = {...exports.default.__vccBase};')
  }
} else {
  if (tempOutput.includes('exports.render = render;')) {
    node.add(';exports.default = {...exports.default, render};')
  } else {
    node.add(';exports.default = {...exports.default};')
  }
}
```
this will have an output of 
```
';exports.default = {...exports.default.__vccBase, render};'
```
or 

```
';exports.default = {...exports.default.__vccBase};'
```

lmiller1990: Right, true, that is very complex. Fair enough.

I wonder if we should create a `isClassComponent` function 🤔 

Anyway, this will do for now, I will test it out a bit and if it's all good do a release. thanks!

PR #302, opened 2020-11-19
Don't require ts-jest and typescript if you're not using typescript

boboldehampsink:
The resulting boolean is already validated at https://github.com/vuejs/vue-jest/blob/next/lib/process.js#L103

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -64,6 +64,9 @@ const getBabelOptions = function loadBabelOptions(filename, options = {}) {
 }
 
 const getTsJestConfig = function getTsJestConfig(config) {
+  if (getVueJestConfig(config).tsConfig === false) {
+    return false;
+  }
   const createTransformer = require('ts-jest').createTransformer
   const tr = createTransformer()
   const configSet = tr.configsFor(config)
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -65,7 +65,7 @@ const getBabelOptions = function loadBabelOptions(filename, options = {}) {
 
 const getTsJestConfig = function getTsJestConfig(config) {
   if (getVueJestConfig(config).tsConfig === false) {
-    return false;
+    return null
   }
   const createTransformer = require('ts-jest').createTransformer
   const tr = createTransformer()


lmiller1990:
Thanks! Seems the linter is failing - can you `yarn lint:fix` (or whatever the command is).

I can release this once it's passing.

boboldehampsink:
@lmiller1990 running `yarn lint:fix` doesn't yield any changes, but I did change `false` to `null` to be more consistent and more correct

boboldehampsink:
Seems that fixed the issue as well

lmiller1990:
Thanks, one last question any reason to do `=== false` instead of the usual `if (!getVueJestConfig(config).tsConfig)) {` ?

boboldehampsink:
@lmiller1990 yeah we want the user to explicitly disable typescript. So this only works when that value is false, not when it is, for example, omitted

lmiller1990:
I think I am confused here. Isn't the current problem that TS is required even when the user isn't using TS in their project? I don't think the user should have to opt out - if they are not using TS (eg, no tsconfig present) then that should be enough to know they are not using TS.

boboldehampsink:
@lmiller1990 agree, but I'm just looking at the docs which say to explicitly disable typescript because tsConfig's default is true: https://github.com/vuejs/vue-jest#boolean-1

lmiller1990:
Interesting that seems like a kind of bad default. I think we should do what makes sense, I doubt most people would think to look for this kind of config option. What do you think? Are you interested in implementing the above (check for a ts-config, if there isn't one, do not use TS)?

boboldehampsink:
Sorry, don't have the time or package knowledge to think about how this should work properly. For now I am already helped if this PR can be merged and released, so that I (and others) can use this without TS at least. You should create a new issue for this, so that a new PR can be submitted.

lmiller1990:
All good, I will add a commit to this PR! Thanks for your help 👍

boboldehampsink:
@lmiller1990 can we get this out in the meanwhile?

lmiller1990:
I am having trouble testing this and verifying we actually solved the problem - I cannot get a repository to throw the error listed in the original issue.

lmiller1990:
What do you think about this solution instead? https://github.com/vuejs/vue-jest/pull/307

lmiller1990:
Closed in favor of #307

PR #307, opened 2020-12-07
fix: do not require ts-jest and typescript

lmiller1990:
We can use the `tsconfig` module to see if a `tsconfig.json` exists. If it doesn't, we can assume they are not using TypeScript. This removes the need to install TS for non TS projects.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,5 +1,6 @@
 const constants = require('./constants')
 const loadPartialConfig = require('@babel/core').loadPartialConfig
+const { resolveSync: resolveTsConfigSync } = require('tsconfig')
 const chalk = require('chalk')
 const path = require('path')
 const fs = require('fs')
@@ -64,6 +65,11 @@ const getBabelOptions = function loadBabelOptions(filename, options = {}) {
 }
 
 const getTsJestConfig = function getTsJestConfig(config) {
+  const isUsingTs = resolveTsConfigSync(process.cwd())
+  if (!isUsingTs) {
+    return null
+  }
+
   const createTransformer = require('ts-jest').createTransformer
   const tr = createTransformer()
   const configSet = tr.configsFor(config)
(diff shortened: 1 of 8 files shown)


boboldehampsink:
This installs TS as a direct dependency?

boboldehampsink (approved):

boboldehampsink:
@lmiller1990 Time for a release? ;-)

lmiller1990:
I can do one, was hoping to look at https://github.com/vuejs/vue-jest/issues/309 first, since we currently assume `tsconfig.json` is at the root dir, which is not always the case.

At the moment you just need to install `ts-jest` to work-around, is this blocking you?

Either way we can probably do a release in the next day or two.

williamweckl:
@lmiller1990 I see that the other issue you pointed to is already solved. Could you release a new version with the both fixes?

Thank you! :)

lmiller1990:
Ok I will do it this week

lmiller1990:
It's out https://github.com/vuejs/vue-jest/releases/tag/v5.0.0-alpha.8

Very tiny release. I hope I can dedicate some time soon to fixing the many TS problems we have. I think it should not be too difficult.

lib/utils.js
```diff
@@ -64,6 +65,11 @@ const getBabelOptions = function loadBabelOptions(filename, options = {}) {
 }
 
 const getTsJestConfig = function getTsJestConfig(config) {
+  const isUsingTs = resolveTsConfigSync(process.cwd())
```
xiaohulu, line 68: Should support tsconfig not place at root dir.
If run `npm run jest`, `process.cwd()` always be root dir. I put tsconfig file at `<rootDir>/sub`.

lmiller1990: Yes, this is a problem. We should support something like

```js
// jest.config.js
module.exports = {  
  globals: {
    'vue-jest': {
      tsconfig: './some/path/tsconfig.json'
    }
  }
}
```

Would you like to make a PR? I can review + release when it's done. If I need to do it, it might take a while, I have a few other projects I need to work on like Test Utils 2.0.0 release etc. I'd really appreciate the help!

PR #315, opened 2021-01-17
add source map dependency (next)

ZachJW34:
`vue-jest` has a [dependency](https://github.com/vuejs/vue-jest/blob/next/lib/map-lines.js#L1) on `source-map` but it is not listed in the package.json. This can cause a resolution issue inside a repo with another package that uses `vue-jest` > 0.7 since there was a [breaking change](https://github.com/mozilla/source-map/blob/master/CHANGELOG.md#070).


lmiller1990:
Cool, thanks

PR #323, opened 2021-03-10
docs: `babelConfig` & `tsConfig` options are removed in v4

sodatea:
As stated in [4.0.0-beta.0 release notes](https://github.com/vuejs/vue-jest/releases/tag/v4.0.0-beta.0):

- To configure TS, `ts-jest` options are expected to be used instead;
- To configure Babel, users should use a custom transformer instead,
(in the same way as a custom block transformer, see https://github.com/vuejs/vue-jest/blob/v4.0.1/e2e/__projects__/custom-transformers/package.json#L42 for example)

I'm not sure where to put these instructions, but we need to remove
the outdated documentation first.


lmiller1990:
Thanks @sodatea ... we sure do, this repo is quite messy and the docs are all over the place.

PR #324, opened 2021-03-10
feat: add support for specific tsconfig file path inside of vue-jest jest.config.js options

xiaohulu:


lib/utils.js
```diff
@@ -65,7 +65,11 @@ const getBabelOptions = function loadBabelOptions(filename, options = {}) {
 }
 
 const getTsJestConfig = function getTsJestConfig(config) {
-  const isUsingTs = resolveTsConfigSync(process.cwd())
+  const tsConfigPath = path.resolve(
+    process.cwd(),
+    config.globals?.['vue-jest']?.tsConfig ?? ''
```
nogic1008, line 70: [Optional Chaining](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Operators/Optional_chaining) is not supported on Node < 14.
```suggestion
    getVueJestConfig(config).tsConfig || ""
```

nogic1008 (commented):
This PR causes linguish error. Please fix it.

PR #331, opened 2021-04-06
Fix coverage reports for typescript for vue3

jolting:
This was based on #330 but applied to Vue3.

This properly generates the coverage reports with the filename TsSrc.ts instead of module.ts
```
-   module.ts               |       75 |      100 |    66.67 |       75 |                20 |
+   TsSrc.ts                |       75 |      100 |    66.67 |       75 |                20 |
 ```


lmiller1990 (approved):
thanks

Issue #332, opened 2021-04-07
export default *some functional component* is broken


0x009922:
Repo: https://github.com/0x009922/vue-jest-default-export-issue

## Problem

Something goes wrong when i am exporting function as default export from .vue file. It is common case when my component is functional.

`MyTest.vue`:
```vue
<script lang="ts">
import { FunctionalComponent, h } from 'vue'

const component: FunctionalComponent = () => h('div', {}, '._.')

// console.log for debug, you will see
console.log('Component in MyTest.vue', component);

export default component;
</script>
```

`MyTest.spec.ts`:
```ts
import {mount } from '@vue/test-utils'
import MyTest from './MyTest.vue'

// for debug, see output
console.log('MyTest in spec.ts', MyTest)

test('test', () => {
    const wrapper = mount(MyTest);

    expect(wrapper.html()).toEqual('<div>._.</div>')
})
```

And when i have run `jest`, i got this output:

```
 FAIL  src/MyTest.spec.ts
  × test (17 ms)

  ● test

    expect(received).toEqual(expected) // deep equality

    Received: "<!---->"
         |                            ^
      10 | })

      at Object.<anonymous> (src/MyTest.spec.ts:9:28)

  console.log
    Component in MyTest.vue [Function: component]

  console.log
    MyTest in spec.ts {}

      at Object.<anonymous> (src/MyTest.spec.ts:4:9)

  console.warn
    [Vue warn]: Component is missing template or render function.
      at <Anonymous ref="VTU_COMPONENT" >
      at <VTUROOT>

      at warn (node_modules/.pnpm/@vue/runtime-core@3.0.9/node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:40:17)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 total
Snapshots:   0 total
Time:        2.306 s
Ran all test suites.
 ERROR  Test failed. See above for more details.
```

In `console.log` results you can see, that `MyTest` component in the `MyTest.spec.ts` is not a function. But it is in the `MyTest.vue`.

lmiller1990:
Thanks for the bug report!

Debugging these is super painful. What I do is `console.log` inside [here](https://github.com/vuejs/vue-jest/blob/next/lib/generate-code.js#L42) and see what's going on. Looks like the render function is not assigned correctly.

I can look at this in the near-ish future, but if you want to take a look, happy to help out with debugging.

0x009922:
[...]

PR #335, opened 2021-05-06
fix: process render function, exported as default #332

0x009922:
I just commented out line `node.add(';exports.default = {...exports.default};')`, added simple test, and everything just works.
Perhaps, in some edge cases, you will still need to copy `exports.default` into yourself. Then you can try this option:

```js
node.add(`;
  if (typeof exports.default === 'function') {
    const fn = exports.default;
    exports.default = (...args) => fn(...args);
    Object.assign(exports.default, fn)
  } else {
    exports.default = {...exports.default};
  }
`)
```


lmiller1990:
Empty else block is a little weird - we can probably remove it. Anyway, Ci is green so I guess this is fine! Thanks.

PR #339, opened 2021-05-10
replace babel-plugin-transform-es2015-modules-commonjs with @babel/pl…

meHimanshu:
Fix to bump up lodash from 4.17.4


lmiller1990:
v3 is not really maintained anymore - I don't plan on doing any more releases.

Are you using it still? Consider upgrading to v4.

PR #349, opened 2021-06-02
add less relative import

YutamaKotaro:
I'm sorry... this issue is related with CSS Modules...  🙇🏻 

### For what ?
#350 

### Why this issue happen ?
We don't have `lang === less` condition in process styles, so I got that issue when I try to use import statement on less file.


lmiller1990 (approved):
Thanks! Will release soon.

lanxisama:
when I use @import in less css module console have some error like
``` javascript
SyntaxError: undefined in input on line 261, column 1:
260 
261 @import (reference) '../../styles/mixin';
262 
```

maybe this commit can resolve this trouble?

PR #360, opened 2021-06-16
build: migrate to monorepo (for vue-jest@4)

nogic1008:
⚠️ **DO NOT merge this PR with "Squash & Merge" method!!** ⚠️
It breaks [package.json](https://github.com/vuejs/vue-jest/blob/master/package.json) histories.

This PR is a part of #355.

## Changes
- setup monorepo (use [yarn workspace](https://classic.yarnpkg.com/en/docs/workspaces/))
- move `e2e/__projects__/*` projects to `e2e/2.x/*`
- move `package.json` and `lib/` to `packages/vue2-jest/`
- change package to `vue2-jest@26.0.0`
- fix e2e test for monorepo

Issue #355, opened 2021-06-14
Migrate to monorepo to split Vue 2 and Vue 3 packages


nogic1008:
Related #343

Currently, the structure of this repository is as follows:
|Branch|Version|Vue|Jest|
|------|--------|----|-----|
|master|4.x|2.x|<27|
|next|5.x(alpha)|3.x|<27|

In Jest 27, breaking changes have came. (#344, #348, #351)
To accommodate it, this package also requires breaking changes. (and need to increment semver)

But, `vue-jest@4` cannot increment because of 5.x exists.
So we should split it to another package.

If you want to keep the package name different, I don't think it is necessary to divide the version by branch. (ex. CI/CD)

@lmiller1990
Could you tell us your thoughts on the items in the "Consideration" section?

### Package naming & Support version

|Package|Ver.|Vue ver.|Jest ver.|Description|
|---------|----|---------|---------|------------|
|`vue-jest`|4.x|2.x|<27|alias for `vue2-jest@26`|
|`vue-jest`|5.x(alpha)|3.x|<27|deprecate (use `vue3-jest@26` instead)|
|`vue-jest`|27.x|latest (=2.x now)|27.x|alias for `vue2-jest@27` now|
|`vue2-jest`|26.x|2.x|26.x|equals to `vue-jest@4`|
|`vue3-jest`|26.x|3.x|26.x|equals to `vue-jest@5`|
|`vue2-jest`|27.x|2.x|27.x||
|`vue3-jest`|27.x|3.x|27.x||

### Versioning

Like [ts-jest](https://github.com/kulshekhar/ts-jest#versioning), We **DOES NOT** use [Semantic Versioning](https://semver.org/).

```
vue2-jest@xx.yy.zz
vue3-jest@xx.yy.zz
```

- `xx`: Equals supported Jest major version
- `yy`: Compliant with semVer major & patch version (ie included **BREAKING CHANGE**)
- `zz`: Compliant with semVer patch version

It is recommended for users to pin version as below:
```jsonc
"devDependencies": {
  "vue2-jest": "^27.1.1", // Not recommended. It means 27.1.1 <= x < 28.0.0
  "vue2-jest": "27.1.x", // 27.1.0 <= x < 27.2.0
  "vue2-jest": "~27.1", // 27.1.0 <= x < 27.2.0
  "vue2-jest": "~27.1.1" // 27.1.1 <= x < 27.2.0
}
```

## Repository Structures

<details>

[...]

PR #361, opened 2021-06-16
build: migrate to monorepo (for vue-jest@5)

nogic1008:
⚠️ **DO NOT merge this PR with "Squash & Merge" method!!** ⚠️
It breaks [package.json](https://github.com/vuejs/vue-jest/blob/next/package.json) histories.

This PR is a part of #355.
Similar to #360, but targeted at `next`(vue-jest@5) branch.

## Changes
- setup monorepo (use [yarn workspace](https://classic.yarnpkg.com/en/docs/workspaces/))
- move `e2e/__projects__/*` projects to `e2e/3.x/*`
- move `package.json` and `lib/` to `packages/vue3-jest/`
- change package to `vue3-jest@26.0.0-alpha.10`
- fix e2e test for monorepo


packages/vue3-jest/package.json
```diff
@@ -0,0 +1,77 @@
+{
+  "name": "vue3-jest",
+  "version": "26.0.0-alpha.10",
```
lmiller1990, line 3: How do you think we should handle this? Should this match a specific Jest version (maybe latest 26.x minor version?)

packages/vue3-jest/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "jest": "^26.0.0",
+    "semantic-release": "^15.13.2",
+    "ts-jest": "^26.4.4",
+    "typescript": "^4.1.2",
+    "vue": "^3.0.3"
+  },
+  "peerDependencies": {
+    "@babel/core": "7.x",
+    "babel-jest": ">= 24 < 27",
+    "jest": ">= 24 < 27 ",
```
lmiller1990, line 41: I think we should only match Jest 26 now.

lmiller1990 (commented):
I left some comments.

I have a question that we did not think about. We will need to support multiple versions of Jest at the same time for a while - in this case, we need to maintain v26 and v27 support.

Should we have two packages? `vue3-jest26` and `vue3-jest27`? What do you think is the best way to handle this?

nogic1008:
@lmiller1990 
I think it's appropriate to separate the branches depending on supported Jest version.

- `master` branch
  - `vue2-jest@27`
  - `vue3-jest@27`
- `26.x` branch
  - `vue2-jest@26`
  - `vue3-jest@26`

packages/vue3-jest/package.json
nogic1008: Changed to `26.x`.

lmiller1990:
@nogic1008 I see. I wonder if there will be any complexity in having both Vue 2 and Vue 3 in the same monorepo? Are we able to do this (I guess we need to do `nohoist`?)

Is there anything else you'd like to do before merging this? Otherwise, I'll merge and do a release to make sure everything is 👌 . Then we can work on getting Jest 27 support merged.

nogic1008:
I believe that the complexity of including different Vue version in `node_modules` is less than the complexity of having to separate the same issues on different branches. (ex. security update by dependabot)
I copied /packages and /e2e from `next` branch to `master` branch locally, but it seems that `nohoist` settings are not needed at this time.

We need to merge `next` branch into `master` branch to complete the monorepo.
README also needs to be added and updated. (in root and each packages)

lmiller1990:
Ok, let's try your idea. 

I think we can update the README later, next goal should be Jest 27.

PR #364, opened 2021-07-10
feat: support custom blocks for Vue 3

kazupon:
~~This is WIP pull request.~~

As the title, I’ll support be transforming custom blocks for Vue 3.
At this time, only jest 26.x is supported.

### DIFF
--- a/packages/vue3-jest/lib/generate-code.js
+++ b/packages/vue3-jest/lib/generate-code.js
@@ -15,13 +15,15 @@ function addToSourceMap(node, result) {
   }
 }
 
-module.exports = function generateCode(
+module.exports = function generateCode({
   scriptResult,
   scriptSetupResult,
   templateResult,
+  customBlocksResult,
+  componentNamespace,
   filename,
   stylesResult
-) {
+}) {
   var node = new SourceNode(null, null, null)
   addToSourceMap(node, scriptResult)
   addToSourceMap(node, scriptSetupResult)
@@ -61,6 +63,14 @@ module.exports = function generateCode(
         mergedStyle
       )}}`
     )
+  }
+
+  if (Array.isArray(customBlocksResult)) {
+    customBlocksResult.forEach(codes => {
+      codes.forEach(code => {
+        node.add(`;((${componentNamespace}) => { ${code} })(exports.default);`)
+      })
+    })
   }
 
   return node.toStringWithSourceMap({ file: filename })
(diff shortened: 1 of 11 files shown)


packages/vue3-jest/lib/generate-code.js
```diff
@@ -63,5 +65,13 @@ module.exports = function generateCode(
     )
   }
 
+  if (Array.isArray(customBlocksResult)) {
+    customBlocksResult.forEach(codes => {
+      codes.forEach(code => {
+        node.add(`;((${componentNamespace}) => { ${code} })(exports.default);`)
```
kazupon, line 71: Custom blocks can now be handled for components like [`vue-loader`](https://github.com/vuejs/vue-loader/blob/5f411122c65071abbb38363c07de636b038b3aff/src/index.ts#L259) and [`@vitejs/plugin-vue`](https://github.com/vitejs/vite/tree/main/packages/plugin-vue#example-for-transforming-custom-blocks).

packages/vue3-jest/lib/process.js
```diff
@@ -133,23 +134,28 @@ function processStyle(styles, filename, config) {
 
 module.exports = function(src, filename, config) {
   const { descriptor } = parse(src, { filename })
+  const componentNamespace =
+    getVueJestConfig(config)['componentNamespace'] || vueComponentNamespace
```
kazupon, line 138: We have added the `componentNamespace` option to the vue-jest config.
This is because we want the transformer (parser) side of custom blocks to be able to handle them flexibly.

In the case of i18n custom blocks, we are using a common library for generic processing for various bundlers such as vue-i18n-loader, rollup-plugin-vue-i18n, and vite-plugin-vue-i18n. For example, it's handled the below:

https://github.com/intlify/cli/blob/4875ec52e0256bf2cf9723dde999b8efa8c043cd/src/generator/json.ts#L91-L103

In the code in the URL above, the component pass to the function, and uses the variable name `Component`.

In order to reuse this logic in vue-jest transformer, I need to change it a bit, but in any case, being able to specify the variable name allows for flexible processing on the transformer side.

lmiller1990:
I will review this and port to Jest 27 in the near future, thanks for waiting.

kazupon:
Hi! @lmiller1990 
I'm working with @nogic1008 and my project member on vue-i18n-jest, which is now monorepo like vue-jest.
https://github.com/intlify/vue-i18n-jest/issues/42

What do we do need additional work?

lmiller1990:
Sorry I took a while to look at this. Seems fine!

Let's release this right now.

kazupon:
Thank you very much! ✨ ❤️

PR #343, opened 2021-05-26
feat: support jest v27 in vue3-jest

cexbrayat:
And also babel-jest and ts-jest v27.

The PR is currently failing because of a change/regression in ts-jest (update: fixed in ts-jest v27.0.1)

TODO:
- [x] wait for https://github.com/kulshekhar/ts-jest/pull/2628 to be merged and released

PR #368, opened 2021-07-25
feat: jest v27 support for vue2-jest

cexbrayat:
Followup, and quasi identical PR to #343 

@lmiller1990 if that looks good, it would be nice to have an alpha release. I think I need it to update the CLI to jest 27 (as it looks like I need to bump Vue 3 and Vue 2 projects at the same time)

### DIFF
--- a/e2e/2.x/babel-in-package/package.json
+++ b/e2e/2.x/babel-in-package/package.json
@@ -16,12 +16,13 @@
     "@babel/preset-env": "^7.9.0",
     "@vue/test-utils": "^1.1.0",
     "coffeescript": "^2.3.2",
-    "jest": "26.x",
-    "ts-jest": "^26.3.0",
+    "jest": "27.x",
+    "ts-jest": "^27.0.4",
     "typescript": "^3.2.2",
-    "vue2-jest": "~26.0.0"
+    "vue2-jest": "~27.0.0-alpha.1"
   },
   "jest": {
+    "testEnvironment": "jsdom",
     "moduleFileExtensions": [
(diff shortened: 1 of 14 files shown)


lmiller1990:
I will do this week, sorry about the delay.

lmiller1990:
Alpha release will be in 24-48h! Thanks for the patience.

lmiller1990:
@cexbrayat just to confirm and make sure we are on the same page, we want this released under `vue-jest@27.0.0-alpha.0`?

`vue2-jest` is not available as you pointed out (I think it was you). I need to email `npm` to get access: https://www.npmjs.com/package/vue2-jest

cexbrayat:
@lmiller1990 No, I don't think we should release it as vue-jest v27. It would be too confusing in my opinion to have vue-jest v4 for vue 2, v5 for vue 3 and v27 for vue 2 again. If we publish vue-jest v27, it should be the same as vue3-jest v27.

Ideally, it would be nice to have `vue2-jest` back from NPM, or maybe we could publish this under `@vue` (`@vue/vue2-jest` and `@vue/vue3-jest`?).

afontcu:
Publishing under `@vue` looks like a valid solution? It makes total sense as it in an official package

lmiller1990:
Ok, I will publish it under `vue` for now (not sure if need any special permissions, I'll give it a shot).

We will need to reach out to npm regarding `vue2-jest`. I will look into this, although it might need to come from a `@vuejs.org` email. Let's see how it goes. I'll report back.

PR #377, opened 2021-08-30
feat: enable sugar ref take 2 for vue3-jest

cexbrayat:
This PR allows testing components using sugar ref take 2 https://github.com/vuejs/rfcs/discussions/369

Note that I did not make it configurable for now, but maybe it should be. Opinions?

### DIFF
--- a/e2e/3.x/babel-in-package/package.json
+++ b/e2e/3.x/babel-in-package/package.json
@@ -7,12 +7,12 @@
     "test": "jest --no-cache test.js"
   },
   "dependencies": {
-    "vue": "^3.0.3"
+    "vue": "^3.2.6"
   },
   "devDependencies": {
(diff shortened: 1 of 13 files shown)


lmiller1990:
Nice. What do you mean by configurable? If you aren't using the ref sugar, you won't be impacted by this update, right?

cexbrayat:
I _think_ you won't be impacted (we could merge and release an alpha to make sure of it on our projects).

The probable edge case is someone naming its own variables `$ref` or another reserved sugar syntax... In that case, vue-jest will transform them and probably error, whereas vue-next will not unless explicitly told? I'm not 100% sure, as I'm not using sugar ref in my own projects.

lmiller1990:
It seems really unlikely someone named their variable `$ref`, so I think this is fine.

PR #380, opened 2021-09-03
feat: add compilerOptions to vue-jest global options

SasanFarrokh:
This PR allows to pass down compilerOptions to vue-jest transformation.

### DIFF
--- a/packages/vue3-jest/lib/process.js
+++ b/packages/vue3-jest/lib/process.js
@@ -93,7 +93,8 @@ function processTemplate(descriptor, filename, config) {
     preprocessOptions: vueJestConfig[template.lang],
     compilerOptions: {
       bindingMetadata: bindings,
-      mode: 'module'
+      mode: 'module',
+      ...vueJestConfig.compilerOptions
     }
   })
 


SasanFarrokh:
Could you review this? @cexbrayat

lmiller1990:
Seems fine to me!

lmiller1990:
What is your use-case, btw?

lmiller1990 (approved):

cexbrayat:
Lachlan already merged, but I'm also curious about the use case.

And I think adding a test would be great, so we can remember why this is there, and not have a regression.

SasanFarrokh:
@lmiller1990 @cexbrayat Hey, Thanks for merging.
About the use case, we're using custom directives and other custom node transformations. Just to mention, one of them is `v-test` directive that transforms to `data-test` in test environments (in production mode it does nothing).
So we needed to pass the compiler options down to vue-jest.
I thought it would be a good option to be able to customize the whole compilerOptions.

cexbrayat:
@SasanFarrokh Sounds cool 👍 

Would you mind opening a PR with a small test for this use case? It can be a simple one, just to make sure we don't have a regression later.

lmiller1990:
A test would be great.

SasanFarrokh:
A Simple test added, just to ensure compilerOptions had been used.
https://github.com/vuejs/vue-jest/pull/385

PR #390, opened 2021-09-30
fix(vue2-jest): interop custom transformer require

### LINKED ISSUE #383: v.27.0.0-alpha.2 - TypeError: Jest: a transform must export a `process` function.
According to the ts-jest team, this way (https://github.com/vuejs/vue-jest/blob/v27.0.0-alpha.2/packages/vue3-jest/lib/index.js#L3) of exporting transformer no longer works with Jest 27.

The right way to export transformer is show here: https://github.com/kulshekhar/ts-jest/blob/master/src/index.ts :

```typescript
import { TsJestTransformer } from './ts-jest-transformer'

export default {
  createTransformer: (): TsJestTransformer => new TsJestTransformer(),
}
```

blake-newman:
interop custom transformer require. Some transformers may not interop correctly when requiring, return `transformer.default` or `transformer`.

@vue/vue3-jest fix: #391
fixes #383

### DIFF
--- a/packages/vue2-jest/lib/utils.js
+++ b/packages/vue2-jest/lib/utils.js
@@ -104,6 +104,7 @@ const getCustomTransformer = function getCustomTransformer(
     require(resolvePath(transformerPath))
   ) {
     transformer = require(resolvePath(transformerPath))
+    transformer = transformer.default || transformer
   } else if (typeof transformerPath === 'object') {
     transformer = transformerPath
   }


nogic1008 (approved):

leonardobenedet (approved):

lmiller1990:
I wonder if we still need this?

blake-newman:
@lmiller1990 i'll mark the PR as draft. It may not be needed but yet to upgrade jest to that version.

PR #391, opened 2021-09-30
fix(vue3-jest): interop custom transformer require

### LINKED ISSUE #383: v.27.0.0-alpha.2 - TypeError: Jest: a transform must export a `process` function.

blake-newman:
interop custom transformer require. Some transformers may not interop correctly when requiring, return `transformer.default` or `transformer`.

@vue/vue2-jest fix: #390
fixes #383

### DIFF
--- a/packages/vue3-jest/lib/utils.js
+++ b/packages/vue3-jest/lib/utils.js
@@ -115,6 +115,7 @@ const getCustomTransformer = function getCustomTransformer(
     require(resolvePath(transformerPath))
   ) {
     transformer = require(resolvePath(transformerPath))
+    transformer = transformer.default || transformer
   } else if (typeof transformerPath === 'object') {
     transformer = transformerPath
   }


cexbrayat (commented):
Thanks for the PR! It would be great to add a test for this. Do you have an example of such a transformer? If so you can add a test in the e2e test directory, it's fairly straightforward. That way we're making sure we fix the proper thing and we will not have regressions in the future.

blake-newman:
Sure. I'll add a test this morning. 

This happens with babel-jest. The use case here is when loading babel-jest for TS transformation. This is done when Babel is used for transformation.

PR #395, opened 2021-10-01
fixed `lang='tsx'` files not working for vue2-jest

Yama-Tomo:
Fix lang='tsx' files to be transpiled as tsx.

`typescript.transpileModule` does not correctly parse JSX syntax when the file extension is not `.tsx`.


lmiller1990:
I can release over this weekend. Thanks!

cameron-martin:
Forcing an ES5 target isn't ideal for us so I like this change. Do you know when you'll do a release with this in?

PR #398, opened 2021-10-08
feat: pass styleOption into compiler

Szymon-dziewonski:
From https://github.com/sass/dart-sass/releases/tag/1.34.0 there are new couple breaking changes e.g `10/2` now should be written as `math.div(10/2)`. With that change there `deprecated` message info which pollutes logs. 
Of course for your own code people can adjust scss/sass/less files, however for external libraries like `bootstrap 4` it's not possible, which means there is a lot of info messages that slows build time. For webpack build we can pass options
`Add a --quiet-deps flag which silences compiler warnings from stylesheets loaded through --load-paths.`
like 
```
loaderOptions: {
    sass: {
        sassOptions: {
           quietDeps: true,
       }
    },
},
```

Thats why I did this PR because its no possible for jest tests to pass additional options such as `quietDeps: true`

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,3 +231,24 @@ You can provide [TemplateCompileOptions](https://github.com/vuejs/component-comp
   }
 }
 ```
+
+## Style options
+
+Possbility to change style loader options (sass, scss, less etc).
+
+`styleOptions`: `Object` Default `{}`.
+
+```json
+{
+  "jest": {
+    "globals": {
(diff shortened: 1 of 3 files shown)


Szymon-dziewonski:
@lmiller1990 I can probably create another PR for v3 version if you would like me to do :)

lmiller1990:
Sure, that'd be great!

Issue #409, opened 2021-11-15
Remove explicit dependency installation of `@vue/compiler-sfc`


Theiaz:
Hello,

I'm not sure if this is even an "issue", but a little code improvement. Currently, I'm on a project using Vue 3 with Vue3-jest:

Some lines of the `package.json`:
```
  "devDependencies": {
    "@babel/core": "7.16.0",
    "@babel/preset-env": "7.16.0",
    "@vue/compiler-sfc": "3.2.22",
    "@vue/test-utils": "2.0.0-rc.16",
    "@vue/vue3-jest": "27.0.0-alpha.3",
    "babel-jest": "27.3.1",
    "jest": "27.3.1",
  },
  "dependencies": {
    "@jest/globals": "27.3.1",
    "vue": "3.2.22",
    "vue-loader": "16.8.3",
  }
```

I stumbled upon following [change](https://github.com/vuejs/vue-next/commit/64aa8e26ae6c5db927b18bde01fbe0eb4ef460aa#diff-dc9ab23fe544d1125e7b5833cbe444a2b3d0176aeceef660d8ffbb2b5df99b1e) with those lines of text: 

> Note: as of 3.2.13+, this package is included as a dependency of the main vue package and can be accessed as vue/compiler-sfc. This means you no longer need to explicitly install this package and ensure its version match that of vue's. Just use the main vue/compiler-sfc deep import instead.

As far as I can tell, it is no longer necessary to specify the dependency `    "@vue/compiler-sfc": "3.2.22",`  in the package.json. So I removed it. However, our tests arent working anymore, jest is throwing following error message:

```
> jest --coverage

Error: Cannot find module '@vue/compiler-sfc'
Require stack:
- /home/node_modules/@vue/vue3-jest/lib/process.js
- /home/node_modules/@vue/vue3-jest/lib/index.js
- /home/node_modules/jest-util/build/requireOrImportModule.js
- /home/node_modules/jest-util/build/index.js
- /home/node_modules/jest-config/build/getCacheDirectory.js
- /home/node_modules/jest-config/build/Defaults.js
- /home/node_modules/jest-config/build/normalize.js
- /home/node_modules/jest-config/build/index.js
- /home/node_modules/jest/node_modules/jest-cli/build/init/index.js
- /home/node_modules/jest/node_modules/jest-cli/build/cli/index.js
- /home/node_modules/jest/node_modules/jest-cli/bin/jest.js
- /home/node_modules/jest/bin/jest.js
[...]

PR #410, opened 2021-11-15
Remove explicit vue-sfc dependency

Theiaz:
It is delivered with Vue by default. 

For more information see  [Issue](https://github.com/vuejs/vue-jest/issues/409)


cexbrayat:
Inaccenditally did the same work in #411 as I needed to update vue3-jest to a more recent Vue version.

I'll close this one, as it needed to be removed in a few extra places (in particular in the Yarn lockfile)

PR #412, opened 2021-11-18
feat: allow to pass compilerOptions via the Jest config for Vue 3

cexbrayat:
These options can be used to define Vue compiler options in `@vue/vue3-jest`.

For example, to enable the newly introduced `propsDestructureTransform` in Vue v3.2.20:

```js
globals: {
  'vue-jest': {
    compilerOptions: {
      propsDestructureTransform: true
    }
  }
}
```

or to disable `refTransform` (which is enabled by default):

```js
globals: {
  'vue-jest': {
    compilerOptions: {
      refTransform: false
    }
  }
}
```

### DIFF
--- a/e2e/3.x/typescript-with-compiler-options/src/components/PropsDestructureTransform.vue
+++ b/e2e/3.x/typescript-with-compiler-options/src/components/PropsDestructureTransform.vue
@@ -1,0 +1,7 @@
+<template>
+  <h1>{{ name }}</h1>
+</template>
+
+<script setup lang="ts">
+const { name = 'name' } = defineProps<{ name?: string }>()
+</script>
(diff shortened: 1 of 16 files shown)


e2e/3.x/typescript-with-compiler-options/src/components/PropsDestructureTransform.vue
```diff
@@ -0,0 +1,7 @@
+<template>
+  <h1>{{ name }}</h1>
+</template>
+
+<script setup lang="ts">
+const { name = 'name' } = defineProps<{ name?: string }>()
```
lmiller1990, line 6: Did not realize this was something now supported in Vue, neat.

lmiller1990 (approved):
Great, thanks!

PR #413, opened 2021-11-18
chore: drop circleci build

cexbrayat:
We have been using Github Actions for quite a while, so the CircleCI job is not really useful anymore (and on outdated node version).


cexbrayat:
We need someone with admin access to CircleCI to stop it from building the project

lmiller1990:
I don't have access I don't think. I will just merge this for now and ping someone to remove it.

Issue #422, opened 2021-11-29
SFC Vue files with tests disappear from test coverage reports


wopian:
I have Jest setup to test files with `@vue/vue3-jest` and it runs all of the Vue tests just fine, however I came across a peculiar issue regarding test coverage - it was going down in our CI dashboard as I was testing more components.

SFC Vue files without a `*.spec.js` test show up in the Jest coverage summary with zero coverage as expected. After adding a `*.spec.js` for a component, it will no longer collect or be listed in the coverage summary in our project.

Regular JavaScript files are unaffected and report coverage as expected.

I have a demo repository setup, however it only partially reproduces our issue - `@components/Demo.vue` is missing, but the two components in `@components/demo/` show up fine (which doesn't happen in our project). Removing the `@components/Demo.spec.js` makes it show up again with zero coverage.

Demo: https://github.com/wopian/jest-coverage-demo

### With Demo.spec.js, coverage unreported
<img width="749" alt="Screenshot 2021-11-29 at 11 45 43" src="https://user-images.githubusercontent.com/3440094/143862489-73cc4042-30f8-4de5-8e44-d7585828018a.png">

### With Demo.spec.js.old, (zero) coverage reported
<img width="760" alt="Screenshot 2021-11-29 at 11 44 19" src="https://user-images.githubusercontent.com/3440094/143862276-0be0b6cc-d02a-4b63-813d-146ce2e66c91.png">

```
vue: ^3.2.22
jest: ^27.3.1
@vue/vue3-jest: ^27.0.0-alpha.4
```

lmiller1990:
Odd, definitely a bug. Is this something you might be interested in looking into? I'd recommend trying to reproduce it in this repo next, then we can easily debug, test and deploy a fix.

wopian:
I've added a more minimal reproduction project there, though I don't know if I will have the free time to investigate further due to work commitments

lucassonsinlima:
Facing the same here, I'm trying to find out something that could cause this but I not went so far

```
vue: ^3.2.26
@vue/cli-plugin-unit-jest: ^5.0.0-rc.1
@vue/vue3-jest: ^27.0.0-alpha.4
```

lmiller1990:
If anyone wants to take a look at this, it would be great! I'm too busy to look at `vue-jest` bugs at this point in time.

jonsalvas:
for those who are interested: we had the same issue in our project and as workaround we returned to 
```
[...]

PR #423, opened 2021-12-03
chore: add missing code coverage e2e for #422

wopian:
See https://github.com/vuejs/vue-jest/issues/422 for details on the issue.

This minimal reproduction reproduces the bug I see in our project, with SFC Vue files that have tests/get imported in `*.spec.js` vanish from coverage reports. Untested SFC Vue files are reported correctly with no code coverage.

Both `Basic.vue` and `Untested.vue` are identical and use the current SFC
`<script setup>` syntax as of Vue 3.2.22

Screenshot 2021-12-03 at 15 14 23

### DIFF
--- a/e2e/3.x/code-coverage/babel.config.js
+++ b/e2e/3.x/code-coverage/babel.config.js
@@ -1,0 +1,13 @@
+module.exports = {
+  presets: [
+    [
+      '@babel/preset-env',
+      {
+        targets: {
+          node: 'current',
+          esmodules: false
+        }
+      }
+    ]
+  ]
+}
--- a/e2e/3.x/code-coverage/components/Basic.spec.js
+++ b/e2e/3.x/code-coverage/components/Basic.spec.js
@@ -1,0 +1,19 @@
+import { assert } from '@vue/compiler-core'
+import { shallowMount } from '@vue/test-utils'
+
+import Basic from './Basic.vue'
+
+describe('Basic', () => {
+  it('renders', () => {
+    expect.assertions(4)
+
+    const wrapper = shallowMount(Basic)
+    const div = wrapper.find('div')
+    const h1 = wrapper.find('h1')
+
+    expect(wrapper.element).toMatchSnapshot()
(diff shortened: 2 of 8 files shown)


wopian:
Should the following be added to `jest.config.js` to deliberately fail the build when coverage from `Basic.vue` is not collected? (it should be somewhere between 25-50% as the method function is unused, but 1% works too)

```js
  coverageThreshold: {
    global: {
      lines: 1
    }
  },
```

wopian:
@lmiller1990 this appears to be an issue between Vue Jest and the default (babel) coverage provider used by Jest.

Running these tests (and in production repo) with the experimental v8 engine as the coverage provider correctly outputs coverage data for SFC Vue files - `--coverageProvider=v8`

lmiller1990:
I see. What's the next step here, then, to get this fixed and merged?

jolting:
Related?
https://github.com/facebook/jest/issues/11776

wopian:
Unfortunately I do not have the available time to investigate why and what package between vue-jest, jest and whichever babel package is responsible for coverage is causing the issue.

> Related? [facebook/jest#11776](https://github.com/facebook/jest/issues/11776)

Checking that issue, it seems to be the exact opposite behaviour of this issue - no (javascript) coverage with v8 and coverage with the default babel coverage provider. 

Here, there's no Vue SFC coverage (javascript is covered fine) with the default babel coverage provider, but v8 coverage provider has Vue SFC coverage.

Roriz:
> @lmiller1990 this appears to be an issue between Vue Jest and the default (babel) coverage provider used by Jest.
> 
> Running these tests (and in production repo) with the experimental v8 engine as the coverage provider correctly outputs coverage data for SFC Vue files - `--coverageProvider=v8`
> 
> Though I'm unable to use that as a temporary client-side fix in GitHub Actions as V8 is 6x slower to run (and without increasing heap size to 4GB, also crashes from OOM)

Thank you for this workaround, it works just fine in my specific case.

Issue #438, opened 2022-01-13
SFC styles are not being applied to component in test environment with VTL


joaopslins:
Repository with reproduction: https://github.com/joaopslins/vue-sfc-styles-issue-repro

I'm using Vue 2 with Vue Testing Library

Using `toBeVisible()` matcher against an element with `display: none;` applied via a class in `<styles>` section does not work in my test. Since I couldn't find other people with the same issue (except [this SO question](https://stackoverflow.com/questions/66535843/sfc-styles-not-being-applied-with-testing-library) where the author solved the issue by other means), I guess this is some configuration issue, or I must be doing something wrong.

I wasn't sure in which repo I should open this issue but I think VTL is working properly, so it must be something related to `vue-jest`. Unless I'm mistaken, I think `vue-jest` should apply the styles correctly, due to https://github.com/vuejs/vue-jest#supported-style-languages.

Component: 
https://github.com/joaopslins/vue-sfc-styles-issue-repro/blob/8d1ac53921dcbe75d81eb3327d07ad9d650344d6/src/HelloWorld.vue#L1-L9

Test:
https://github.com/joaopslins/vue-sfc-styles-issue-repro/blob/8d1ac53921dcbe75d81eb3327d07ad9d650344d6/src/__tests__/HelloWorld.spec.js#L1-L13

Relevant versions:
Node: 14
"jest": "^27.4.7",
"vue": "^2.6.11"
"@vue/vue2-jest": "^27.0.0-alpha.4",

afontcu:
Hi! And thank you for the report.

Looks like you might be right, because changing to inline styles makes the test pass (at least for me locally):

```vue
<template>
  <div class="test" style="display: none">Test!</div>
</template>
```

So it is def something about vue-jest's ability to compile the style block rather than a Vue Testing Library/Vue Test Utils/jest-dom issue.

I don't have the bandwidth right now to find the root cause, but submitting a PR with a failing test (probably [here](https://github.com/vuejs/vue-jest/blob/master/e2e/2.x/style/test.js)?) would help a ton 🙏 I'll look into it at some point.


Thanks!

joaopslins:
Hi @afontcu, I managed to get this PR #439, I'm not sure if this is the reproduction case you're looking for, so please let me know if there's something else that I can/should do, so I can improve the PR. Thanks again for your quick answer!

afontcu:
Hi! Looks like a good starting point, I'll take a look and see what I can come up with. Thanks!

[...]

PR #439, opened 2022-01-14
Adds style failing test case for Vue Testing Library

joaopslins:
Test case reproducing issue #438.

From master, I ran `yarn add @testing-library/vue --dev`, and also had to bump `@vue/test-utils` to `1.3.0` otherwise it wouldn't work. I probably messed up something in the `yarn.lock` file so apologies for that.

I'm not sure if this is the reproducible case you were looking for so if there's something else I can/should do please let me know so I can improve this PR. Thank you!


lmiller1990:
Hi! Sorry for the slow reply. Can you try merging in master and pushing to see if CI passes?

joaopslins:
@lmiller1990 Thanks for the reply. I've force-pushed pulling the changes from master.
Looks like CI has some issue with setup maybe?
Let me know how I can help.

lmiller1990:
I enabled CI for this PR, let's see how it goes.

joaopslins:
I think the test is still failing...

In any case, could you double check if my test is testing the right thing? As in, it should work that way?

PR #486, opened 2022-07-22
Fix source maps for vue2-jest

NikhilVerma:
The code has been borrowed from vue3-jest

For me this fixes: https://github.com/vuejs/vue-jest/issues/474

PR #483, opened 2022-07-05
feat: support <script setup> in Vue 2.7

FelixGraf:
This PR:

- updates all Vue 2 dependencies (`vue` and `vue-template-compiler`) to the latest version (`2.7.7`)
- adds support for parsing components using the `<script setup>` syntax in Vue 2.7 components
  - most of the code (and e2e test) was just copied from the `vue3-jest` package (https://github.com/vuejs/vue-jest/blob/master/packages/vue3-jest/lib/process.js#L51)

PR #489, opened 2022-08-18
feat: support <script setup> in Vue 2.7 (supersedes #483)

kiroushi:
This PR basically borrows @FelixGraf work in #483 and merges snapshot fixes from #486 into it.

The source map in script setup generated code didn’t have a trailing `\n`, so the `__options__` var declaration was being inlined with it.

I removed the source map from the generated Vue 2 code, I don’t know if this is a necessary change, and someone with more insight on the library might shed some light.

### DIFF
--- a/e2e/2.x/basic/__snapshots__/test.js.snap
+++ b/e2e/2.x/basic/__snapshots__/test.js.snap
@@ -2,7 +2,7 @@
 
 exports[`generates source maps for .vue files 1`] = `
 Object {
-  "file": "/home/runner/work/vue-jest/vue-jest/e2e/2.x/basic/components/Basic.vue",
+  "file": "./components/Basic.vue",
   "mappings": ";;;;;;eACe;AACbA,MAAI,EAAE,OADO;AAEbC,UAAQ,EAAE;AACRC,kBAAc,EAAE,SAASA,cAAT,GAA0B;AACxC,aAAO;AACLC,WAAG,EAAE,KAAKC,OADL;AAELC,YAAI,EAAE,CAAC,KAAKD,OAFP;AAGLE,cAAM,EAAE,KAAKF;AAHR,OAAP;AAKD;AAPO,GAFG;AAWbG,MAAI,EAAE,SAASA,IAAT,GAAgB;AACpB,WAAO;AACLC,SAAG,EAAE,4BADA;AAELJ,aAAO,EAAE;AAFJ,KAAP;AAID,GAhBY;AAiBbK,SAAO,EAAE;AACPC,eAAW,EAAE,SAASA,WAAT,GAAuB;AAClC,WAAKN,OAAL,GAAe,CAAC,KAAKA,OAArB;AACD;AAHM;AAjBI,C",
   "names": Array [
     "name",
--- a/e2e/2.x/basic/__snapshots__/test.js.snap
+++ b/e2e/2.x/basic/__snapshots__/test.js.snap
@@ -1,10 +1,10 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`generates source maps for .vue files 1`] = `
-Object {
+{
   "file": "./components/Basic.vue",
   "mappings": ";;;;;;eACe;AACbA,MAAI,EAAE,OADO;AAEbC,UAAQ,EAAE;AACRC,kBAAc,EAAE,SAASA,cAAT,GAA0B;AACxC,aAAO;AACLC,WAAG,EAAE,KAAKC,OADL;AAELC,YAAI,EAAE,CAAC,KAAKD,OAFP;AAGLE,cAAM,EAAE,KAAKF;AAHR,OAAP;AAKD;AAPO,GAFG;AAWbG,MAAI,EAAE,SAASA,IAAT,GAAgB;AACpB,WAAO;AACLC,SAAG,EAAE,4BADA;AAELJ,aAAO,EAAE;AAFJ,KAAP;AAID,GAhBY;AAiBbK,SAAO,EAAE;AACPC,eAAW,EAAE,SAASA,WAAT,GAAuB;AAClC,WAAKN,OAAL,GAAe,CAAC,KAAKA,OAArB;AACD;AAHM;AAjBI,C",
-  "names": Array [
+  "names": [
     "name",
     "computed",
     "headingClasses",
@@ -20,7 +20,7 @@ Object {
   "sources": [
     "Basic.vue",
   ],
-  "sourcesContent": Array [
+  "sourcesContent": [
     "
 export default {
   name: 'basic',
--- a/e2e/2.x/basic/__snapshots__/test.js.snap
+++ b/e2e/2.x/basic/__snapshots__/test.js.snap
@@ -1,10 +1,10 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`generates source maps for .vue files 1`] = `
-{
-  "file": "./components/Basic.vue",
-  "mappings": ";;;;;;AAAA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;eAEe;AACbA,MAAI,EAAE,OADO;AAEbC,UAAQ,EAAE;AACRC,kBAAc,EAAE,SAASA,cAAT,GAA0B;AACxC,aAAO;AACLC,WAAG,EAAE,KAAKC,OADL;AAELC,YAAI,EAAE,CAAC,KAAKD,OAFP;AAGLE,cAAM,EAAE,KAAKF;AAHR,OAAP;AAKD;AAPO,GAFG;AAWbG,MAAI,EAAE,SAASA,IAAT,GAAgB;AACpB,WAAO;AACLC,SAAG,EAAE,4BADA;AAELJ,aAAO,EAAE;AAFJ,KAAP;AAID,GAhBY;AAiBbK,SAAO,EAAE;AACPC,eAAW,EAAE,SAASA,WAAT,GAAuB;AAClC,WAAKN,OAAL,GAAe,CAAC,KAAKA,OAArB;AACD;AAHM;AAjBI,C",
-  "names": [
+Object {
+  "file": "/home/runner/work/vue-jest/vue-jest/e2e/2.x/basic/components/Basic.vue",
+  "mappings": ";;;;;;eACe;AACbA,MAAI,EAAE,OADO;AAEbC,UAAQ,EAAE;AACRC,kBAAc,EAAE,SAASA,cAAT,GAA0B;AACxC,aAAO;AACLC,WAAG,EAAE,KAAKC,OADL;AAELC,YAAI,EAAE,CAAC,KAAKD,OAFP;AAGLE,cAAM,EAAE,KAAKF;AAHR,OAAP;AAKD;AAPO,GAFG;AAWbG,MAAI,EAAE,SAASA,IAAT,GAAgB;AACpB,WAAO;AACLC,SAAG,EAAE,4BADA;AAELJ,aAAO,EAAE;AAFJ,KAAP;AAID,GAhBY;AAiBbK,SAAO,EAAE;AACPC,eAAW,EAAE,SAASA,WAAT,GAAuB;AAClC,WAAKN,OAAL,GAAe,CAAC,KAAKA,OAArB;AACD;AAHM;AAjBI,C",
+  "names": Array [
     "name",
     "computed",
     "headingClasses",
@@ -20,30 +20,8 @@ exports[`generates source maps for .vue files 1`] = `
   "sources": [
     "Basic.vue",
   ],
-  "sourcesContent": [
-    "//
-//
-//
-//
-//
-//
-//
-//
-//
-//
-//
-//
-//
-//
-//
-//
-//
-//
-//
-//
-//
-//
-
+  "sourcesContent": Array [
+    "
 export default {
   name: 'basic',
   computed: {
--- a/packages/vue2-jest/lib/process-custom-blocks.js
+++ b/packages/vue2-jest/lib/process-custom-blocks.js
@@ -1,5 +1,4 @@
 const { getVueJestConfig, getCustomTransformer } = require('./utils')
-const vueOptionsNamespace = require('./constants').vueOptionsNamespace
 
 function applyTransformer(
   transformer,
@@ -17,7 +16,7 @@ function groupByType(acc, block) {
   return acc
 }
 
-module.exports = function(allBlocks, filename, config) {
+module.exports = function(allBlocks, filename, componentNamespace, config) {
   const blocksByType = allBlocks.reduce(groupByType, {})
   const code = []
   for (const [type, blocks] of Object.entries(blocksByType)) {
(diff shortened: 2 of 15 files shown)


ndom91:
Great stuff! Works for me. This was breaking our `<setup script>` + Vue 2 related builds as well.

lmiller1990:
Nice, if CI is green I can merge and release 28.1.1.

lmiller1990:
Looks like we've got some failures - can you take a look @kiroushi?

FelixGraf:
Thanks for taking over my changes, I haven't had the time to continue them. One thing that was mentioned in my PR is [this comment](https://github.com/vuejs/vue-jest/pull/483#discussion_r923730504):

> @vue/compiler-sfc isn't listed as a dependency of this package, so this import is unsound. There is a similar existing issue for the vue-template-compiler import as well (which should probably have been listed as an optional peer dependency).

People using Vue < 2.7 won't have the `@vue/compiler-sfc` dependency installed by default, as it's only available for Vue >= 2.7. This might be a breaking change, as users with older Vue versions should still be able to use `vue-jest` in a new patch/minor version. 🤔

kiroushi:
> People using Vue < 2.7 won't have the @vue/compiler-sfc dependency installed by default, as it's only available for Vue >= 2.7. This might be a breaking change, as users with older Vue versions should still be able to use vue-jest in a new patch/minor version. 🤔

You’re absolutely right. I will spend more time this weekend so we can make this back/forth compatible 👍

kiroushi:
@lmiller1990 can you approve the workflow and see if it’s passing now? I will work on trying to make this backwards compatible too.

dten:
for me 28.1.0 fixed the source maps and this MR breaks them again

lmiller1990:
CI is passing. Hmm - I wonder if @dten can share his setup - we ideally don't want to merge and release a fix that breaks the previous release of 28.1.0. 🤔

dten:
In away from PC got a week now 🙈 if it's got tests and working for everyone else then go for it. There's a chance I messed up testing the pr locally

kiroushi:
@lmiller1990 I didn’t have time to look into the Vue 2.6 vs Vue 2.7 issue. As @FelixGraf stated, these changes will break compatibility with any Vue version not including 2.7 `@vue/compiler-sfc`.

kiroushi:
@dten @lmiller1990 I think this should be good to go now.

I’ve cloned https://github.com/morellexf26/tuto, changed Vue version to `2.6.14`, and modified the test to take a snapshot.

Then I:
- Changed `@vue/vue2-jest` to `28.1.0` and added `"jest-environment-jsdom": "28.0.2"`. Ran test and was passing.
- Changed `@vue/vue2-jest` to this branch. Ran test and was passing (snapshots were OK).
- Changed `vue` dependency to `2.7.10` and changed `SpotifyTracker.vue` component to setup syntax. Ran tests and they were passing, snapshots were still OK.

I think I’ve covered all the cases, let me know if you find an edge case.

cc @FelixGraf too

dten:
That's with `--no-cache` to jest? For me misleading results because it was caching the old maps

packages/vue2-jest/lib/process.js
```diff
@@ -8,8 +8,18 @@ const stripInlineSourceMap = require('./utils').stripInlineSourceMap
...
+
+let isVue27 = false
+let compilerUtils
+
+try {
+  compilerUtils = require('@vue/compiler-sfc')
+  isVue27 = true
+} catch (e) {
+  compilerUtils = require('@vue/component-compiler-utils')
+}
```
thebanjomatic, line 22: Should these both be listed as optional peer dependencies?

packages/vue2-jest/lib/process.js
```diff
@@ -8,8 +8,18 @@ const stripInlineSourceMap = require('./utils').stripInlineSourceMap
 const getCustomTransformer = require('./utils').getCustomTransformer
 const loadSrc = require('./utils').loadSrc
 const babelTransformer = require('babel-jest').default
-const compilerUtils = require('@vue/component-compiler-utils')
```
kiroushi, line 11: @thebanjomatic this import used to be here anyway, without being declared explicitly in the peer dependencies. My exposure to this codebase is limited so I am not really aware of the consequences of these changes given the amount of permutations we can have in regards of `vue-jest`,`jest`, and `vue` versions.

thebanjomatic: The thing is that its already a peer dependency, the project just isn't declaring it in the package.json manifest. This would have issues if you were using yarn pnp for example because it is trying to import a package that isn't declared and you would need to use packageExtensions to inject those declarations as consumers of the library.

If you don't want to make the change in this PR, that's fine. I might pursue a separate PR later to clean up these undeclared peerDependencies and to fully optionalize vue-template-compiler so that you don't need to install that separately for vue 2.7 also.

I do think `vue/compiler-sfc` is still the better import than `@vue/compiler-sfc` though since we already have the vue peer dependency declared and this appears to be the preferred way to import compiler-sfc now that it ships with the vue dependency out of the box.

kiroushi: @thebanjomatic that’s a fair point. I can change this PR. Will update it later.

lmiller1990: LMK when it's ready for a re-review.

Providing compat for all versions is really hard and tedious... :|

lmiller1990: PS just released v29 for Jest 29

You might need to rebase. If this merges, the changes will be in v29 - we can back port to v28, but it's a lot of work to maintain multiple versions, I generally ask people to update to the latest Jest.

 https://github.com/vuejs/vue-jest/releases/tag/v29.0.0

kiroushi: @lmiller1990 sorry I’ve been kinda busy last week, so what’s your recommendation? Make this change only compatible with Jest 29?

kaizige10:
I have tried this PR in my local project. Most of the time it's OK. Except when I use both <script> and  <script setup> in one component to define component name, I find the <script setup> do not execute. 

My component is:
```js
<script>
console.log('in <script>');
export default {
    name: 'LiveHeader',
};
</script>

<script setup>
console.log('in <script setup>');
</script>
```

jest log is:
```
  console.log
    in <script>

      at Object.log (vpages/monitor/components/LiveHeader.vue:2:9)

 FAIL  test/monitor/liveheader.test.js
  LiveHeader
    × check title (5 ms)
    × check popular and onlineUser (4 ms)
```

You will see only `in <script>` has been logged. So I think the <script setup> will not take effect when I use <script>. And I don't know why. Pls check. @kiroushi @FelixGraf

kiroushi:
@thebanjomatic I’ve changed the import to `vue/compiler-sfc`, as per recommended in the package docs itself.

I’ve left out the [peer dependencies](https://github.com/vuejs/component-compiler-utils#why-isnt-vue-template-compiler-a-peerdependency) though; in any case I think that escapes the initial purpose of this PR.

@kaizige10 thanks for reporting that. I’ve fixed the use of combined `script` and `script setup`.

@lmiller1990 the branch is rebased and would require Jest v29 now. I don’t think it would be worth to maintain yet another version combination.

This PR should be ready for shipping now.

e2e/2.x/basic/__snapshots__/test.js.snap
```diff
@@ -1,10 +1,10 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`generates source maps for .vue files 1`] = `
-{
-  "file": "./components/Basic.vue",
-  "mappings": ";;;;;;AAAA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;AACA;eAEe;AACbA,MAAI,EAAE,OADO;AAEbC,UAAQ,EAAE;AACRC,kBAAc,EAAE,SAASA,cAAT,GAA0B;AACxC,aAAO;AACLC,WAAG,EAAE,KAAKC,OADL;AAELC,YAAI,EAAE,CAAC,KAAKD,OAFP;AAGLE,cAAM,EAAE,KAAKF;AAHR,OAAP;AAKD;AAPO,GAFG;AAWbG,MAAI,EAAE,SAASA,IAAT,GAAgB;AACpB,WAAO;AACLC,SAAG,EAAE,4BADA;AAELJ,aAAO,EAAE;AAFJ,KAAP;AAID,GAhBY;AAiBbK,SAAO,EAAE;AACPC,eAAW,EAAE,SAASA,WAAT,GAAuB;AAClC,WAAKN,OAAL,GAAe,CAAC,KAAKA,OAArB;AACD;AAHM;AAjBI,C",
-  "names": [
+Object {
+  "file": "/home/runner/work/vue-jest/vue-jest/e2e/2.x/basic/components/Basic.vue",
```
thebanjomatic, line 5: It seems weird for this to have switched to an absolute path. I would expect this to still be a relative path, but its possible this doesn't really matter much since the source-maps are then consumed by jest on the same machine. Possibly there could be issues if you are sharing a jest cache? Ultimately I don't know enough about this to make a judgement call.

thebanjomatic (approved):
I tested this in a project of mine and it appears to be working correctly, and the concerns I raised earlier have all been addressed. From my perspective this looks good to me, but I did have one question still regarding the sourcemaps, but I wasn't sure how to test that.

e2e/2.x/basic/__snapshots__/test.js.snap
kiroushi: Thanks @thebanjomatic. I don’t think the file property should be problematic, but I will revert it back to the original. I don’t know at which point it did change (because that’s not even the path of my local env, so I am guessing I took over these changes from Felix too).

aldarund:
It breaks script setup with @nuxtjs/composition-api in a new different way than general problem before this pr. With this version it ends up with.
SyntaxError: Cannot use import statement outside a module
That happens if just do something like 
```
<script lang="ts" setup>
import { useAsync } from '@nuxtjs/composition-api'
useAsync(() => {
  const k = 3
})
</script>
```
Which works without this pr.

lmiller1990:
Sorry for going MIA, some other things came up that are more pressing than OSS work for me, but I want to get this out, let's do it. Looks like this has been tested decently well by this point.

Last thing we might want to do is document @aldarund's findings somewhere for Nuxt users?

lmiller1990 (approved):

lmiller1990 (approved):

yarn.lock
```diff
@@ -10639,7 +10677,7 @@ vue-template-es2015-compiler@^1.9.0:
 "vue2-sass-importer-lib@file:e2e/2.x/sass-importer/lib":
   version "1.0.0"
   dependencies:
-    vue2-sass-importer-sass-lib "file:../../.cache/yarn/v6/npm-vue2-sass-importer-lib-1.0.0-1087f856-8699-4ac6-a0fa-2288988256c2-1656107462461/node_modules/sass-lib-v1"
+    vue2-sass-importer-sass-lib "file:../../Library/Caches/Yarn/v6/npm-vue2-sass-importer-lib-1.0.0-7ea9c798-4187-4843-8438-890703751297-1658071612744/node_modules/sass-lib-v1"
```
lmiller1990, line 10680: Weird. looks MacOS specific

lmiller1990:
Actually let's ship it, docs can come later - best unblock the PR and get this out in the wild for some real usage to see if we've broken anything. Most people should pin their vue-jest version so I don't think this should be disruptive.

lmiller1990:
hmm I guess something somewhere changed the snapshot serializing for arrays slightly: https://github.com/vuejs/vue-jest/pull/489/commits/4e93580aa3f7ee5de96b9096bd3e52a5a187e52d

Looks good now, let's ship it.

lmiller1990:
Shipped https://github.com/vuejs/vue-jest/releases/tag/v29.1.0, nice job everyone.

dennybiasiolli:
> Actually let's ship it, docs can come later - best unblock the PR and get this out in the wild for some real usage to see if we've broken anything. Most people should pin their vue-jest version so I don't think this should be disruptive.
> 
> Looks like CI is failing. I'll take look.

Thanks everyone for the update, now it's working (almost) fine.
I just noticed a little error during jest tests when trying to spy on a component method like in this example:

```js
const wrapper = shallowMount(CounterLocal)
const spyIncrement = jest.spyOn(wrapper.vm, 'increment')
wrapper.find('button').trigger('click')
expect(spyIncrement).toHaveBeenCalled()
```

Tests are breaking with `Cannot spy the increment property because it is not a function; undefined given instead`

I found a workaround with

```diff
- const spyIncrement = jest.spyOn(wrapper.vm, 'increment')
+ const spyIncrement = jest.spyOn(wrapper.vm._setupState, 'increment')
```

Example repository here: https://github.com/dennybiasiolli/vue-migration-2-to-3, workaround applied to component
[`tests/unit/views/CounterLocal.spec.js`](https://github.com/dennybiasiolli/vue-migration-2-to-3/blob/main/tests/unit/views/CounterLocal.spec.js#L12)

thebanjomatic:
@dennybiasiolli that seems like it is working as expected with script setup. Components are closed by default. If you need to unit test increment, you should expose it using the `defineExpose({increment})` macro.

dennybiasiolli:
@thebanjomatic thanks, you are right about that. I tried with `defineExpose({increment})` but no luck in spying on the method. It seems to call something else.
I think I need to find another way to test for buttons click and methods call 😄

thebanjomatic:
Ah yes.... I can reproduce, and I think your workaround makes sense. Ultimately the code in `<script setup>` exists on and is called from a different object (_setupState). When you use `defineExpose`, it provides an alias to those functions on the vm. Spying on the vm is only going to rewrite that alias and all the internal references to that function within script setup will still go directly to that implementation instead of your spy.

lmiller1990:
I would *really* recommending finding an alternative testing strategy here. After years of triaging Test Utils and Jest, I learned

- testing implementation details is brittle
- internals can (and do) change in Vue (even minors)
- coupling to testing framework (Jest) is also not ideal

In your example you could do something like

```ts
const wrapper = mount(Component)
await wrapper.find('button').trigger('click')
expect(wrapper.find('p').text()).toEqual('Count is 1')
```

I wouldn't use `p` (brittle, changing to div breaks your test, despite no functional different to user). Consider something better like a `data-test` or `role` selector.

dennybiasiolli:
Thank you @lmiller1990 , your example is correct and it follows the vuejs guidelines about testing:

> When testing, remember to test what a component does, not how it does it.

I need to rethink a few test cases, but it was totally a bad habit of mine.

BTW the repo I linked is an example I will use in a talk at VueDay Verona explaining how to upgrade from Vue 2.6 to Vue 3.x passing through 2.7. Your recent update (29.1) is helping a lot in this process, thank you very much for this update!

lmiller1990:
No problem, but don't thank me, most of the heavy lifting for this was done by the others in this PR :)

packages/vue2-jest/lib/process-custom-blocks.js
```diff
@@ -29,7 +28,7 @@ module.exports = function(allBlocks, filename, config) {
       const codeStr = applyTransformer(
         transformer,
         blocks,
-        vueOptionsNamespace,
+        componentNamespace,
```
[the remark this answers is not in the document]
lmiller1990: Argh, sorry. Can you file an issue with a minimal repro?

Bonus points if you want to make a PR!!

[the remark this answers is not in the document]
lmiller1990: Edit: you already made a PR!

Issue #474, opened 2022-05-20
Source maps for vue2 SFCs using TS are not created/wrongly created for v28


phloose:
After upgrading to Jest v28 SFCs that are using TypeScript and the Composition-API plugin for Vue2 are producing a weird error:

```shell
FAIL  ***/FormDateInput.spec.ts
  ● Test suite failed to run

    No element indexed by 61

      4 | import Vue from 'vue';
      5 |
    > 6 | import FormDateInput from './FormDateInput.vue';
        | ^
      7 |
      8 | const localVue = createLocalVue();
      9 |

      at ArraySet_at [as at] (node_modules/source-map/lib/array-set.js:92:9)
      at BasicSourceMapConsumer.SourceMapConsumer_originalPositionFor [as originalPositionFor] (node_modules/source-map/lib/source-map-consumer.js:626:30)
      at node_modules/@vue/vue2-jest/lib/generate-source-map.js:29:28
          at Array.forEach (<anonymous>)
      at generateSourceMap (node_modules/@vue/vue2-jest/lib/generate-source-map.js:26:38)
      at Object.module.exports [as process] (node_modules/@vue/vue2-jest/lib/process.js:129:15)
      at ScriptTransformer.transformSource (node_modules/@jest/transform/build/ScriptTransformer.js:619:31)
      at ScriptTransformer._transformAndBuildScript (node_modules/@jest/transform/build/ScriptTransformer.js:765:40)
      at ScriptTransformer.transform (node_modules/@jest/transform/build/ScriptTransformer.js:822:19)
      at Object.<anonymous> (***/FormDateInput.spec.ts:6:1)
```

It seems to me, that vue2-jest cannot produce source-maps or creates them i a way that is not consumable by the underlying source-map lib.

Relevant packages installed in a node 16 environment with npm 8 are:
- "@vue/vue2-jest": "^28.0.0"
- "jest": "^28.1.0"
- "ts-jest": "^28.0.2"
- "typescript": "^4.6.3"

I know that these version do not match exactly, but it makes no difference: Even if i downgrade all to exactly v28.0.0 it does not work

The transformers property lists `@vue/vue2-jest` correctly for transforming vue files, which also worked for vue-jest v27 and jest 27. I have also tried to include the full path in node modules like `<rootDir>/node_modules/@vue/vue2-jest`, but without success.

```javascript
  transform: {
    '.*\\.(vue)$': '@vue/vue2-jest',
[...]

PR #492, opened 2022-08-22
Don't set inputSourceMap on babel-jest transform

josephlbarnett:
When babel-jest merges the inputSourceMap and the outputSourceMap
it appears to remove names from the resulting sourceMap that are
still referenced by elements of the source map.  Not passing the
inputSourceMap appears to solve #474 and still provide source
coverage, but someone who knows how this is supposed to work
should look more closely.

### DIFF
--- a/packages/vue2-jest/lib/transformers/typescript.js
+++ b/packages/vue2-jest/lib/transformers/typescript.js
@@ -21,13 +21,11 @@ module.exports = scriptLang => ({
 
     res.outputText = stripInlineSourceMap(res.outputText)
 
-    const inputSourceMap = res.sourceMapText && JSON.parse(res.sourceMapText)
-
     const customTransformer =
       getCustomTransformer(vueJestConfig['transform'], 'js') || {}
     const transformer = customTransformer.process
       ? customTransformer
-      : babelJest.createTransformer({ inputSourceMap })
+      : babelJest.createTransformer()
 
     return transformer.process(res.outputText, filePath, config)
   }


josephlbarnett:
Pretty sure this isn't want we want to do as we'll lose the mapping chain from vue -> ts -> jest, but here's what I'm seeing:

typescript.transpileModule gives a sourceMap that looks like:

lmiller1990:
Will try get some 👀 on this - we've got a lot of weird incompatible stuff around source maps, the core issue(s) are not super clear.

josephlbarnett:
another option that potentially works is to assume that babel is set up to handle typescript, and just do a single transform instead of transpiling the typescript and then passing to babel-jest:

so transformers/typescript.js becomes:
```
module.exports = scriptLang => ({
  process(scriptContent, filePath, config) {
    // instead of calling typescript compiler directly, rely on babel having typescript set up
    // and just pass to babel-jest with a .ts extension
    const vueJestConfig = getVueJestConfig(config)

    const customTransformer =
      getCustomTransformer(vueJestConfig['transform'], 'js') || {}
    const transformer = customTransformer.process
      ? customTransformer
      : babelJest.createTransformer()

    return transformer.process(scriptContent, filePath + (scriptLang === 'tsx' ? '.tsx' : '.ts'), config)
  }
})
```

I'm not sure if that babel assumption is safe to make in general for vue/typescript projects, but it does appear to work in my project

lmiller1990:
Hi, I don't have time to prioritise this - I normally can just merge + release things that have a test, do we not need one for this? I wonder if we even have some tests around source maps?

You might need to rebase - we released v29 just now: https://github.com/vuejs/vue-jest/releases/tag/v29.0.0

josephlbarnett:
Hi thanks for the update -- I'll look at rebasing and adding a test if I can.  Opened this more as an ask for guidance than an actual suggested fix, but I'll spend some more time trying to narrow this down to something I'd actually recommend as a fix (probably one of the single-transform solutions?).

josephlbarnett:
hmm everything seems to work ok in my project now with 29.x... not sure what fixed it but :shrug:

PR #501, opened 2022-09-22
fix: Revert vueOptionsNamespace changes from <script setup> commit

wolfgangwalther:
As mentioned in https://github.com/vuejs/vue-jest/pull/489#discussion_r977386910, vue2-jest v29.1.0 broke custom block processors.

This PR just reverts a few changes of #489, which look unrelated to the PR and are entirely broken. There seemed to be a `componentNamespace` config option added - but this was documented nowhere and the default value for it was broken.


lmiller1990:
Can you please add a test so we don’t break it again?

thanks!

lmiller1990:
Looks good! It's out: https://github.com/vuejs/vue-jest/releases/tag/v29.1.1

Issue #495, opened 2022-08-29
New tsconfig loading feature doesn't respect tsconfig extends


RhodeHummel:
The new tsconfig loading feature implemented in #471 doesn't properly support extended tsconfig files. For example take the following config for our test files that extends our base config:

_tests/tsconfig.json_
```json5
{
  "extends": "../tsconfig.json",
  "include": [
    "./**/*.ts"
  ],
}
```

_tsconfig.json_
```json5
{
  "compilerOptions": {
    /* shared compiler options set here */
  }
  "include": {
    "src/**/*.ts",
    "src/**/*.vue"
  }
}
```

When the tests are run the config that is loaded and passed to the typescript compiler doesn't include any of the options set in the base config even though `tests/tsconfig.json` properly extends it. Note this was working previously before this PR was merged.

It looks like this was an issue in the past as well but was fixed and now has appeared again.
https://github.com/vuejs/vue-jest/issues/118
https://github.com/vuejs/vue-jest/pull/139

christianopaets:
I found a strange behaviour with tsconfig too.

I've added to jest config my `tsconfig.spec.json` that extends base `tsconfig.json`, and it works as I expected
```js
globals: {
  'vue-jest': {
    tsConfig: './tsconfig.spec.json'
  }
}
```

Rocik:
I can confirm that as well.
Works with what @christianopaets showed together with not using `extends`.
It has been introduced by #471. The code that loads tsconfig is reading just the `compilerOptions` field.

thebanjomatic:
I have also run into this same problem and wound up implementing my own vue-ts-transformer to override the vue-jest default handling for ts script blocks. My version is just using `typescript.getParsedCommandLineOfConfigFile` from the typescript library itself to load the config file, and passing the parsed config into `typescript.transpileModule`

I can see about putting up a PR to make that change upstream.

PR #471, opened 2022-05-11
feat: load TypeScript config from `tsconfig` package instead of `ts-jest`

nogic1008:
**:warning:This PR contains BREAKING CHANGES!!:warning:**

This PR conflicts with #468, so I set as draft.
Resolve #171, #431

## Changes
- Use `tsconfig` package to load TypeScript config.
- Remove `ts-jest` peerDependencies.

### Breaking Changes
- Some tests will break that depend on `ts-jest` config.

```javascript
// jest.config.js
module.exports = {
  globals: {
    "ts-jest": {...} // Not loaded during transforming Vue (use tsconfig.json instead)
  }
}
```

PR #502, opened 2022-10-11
fix(tsconfig): parse extended tsconfigs when transpiling script blocks

thebanjomatic:
A change introduced in v28.1.0 in PR #471 unintentionally changed the behavior of the tsconfig parsing such that configs using "extends" were no longer being considered.

Fixes: #495

### DIFF
--- a/e2e/2.x/basic/components/ExtendedTsConfig.vue
+++ b/e2e/2.x/basic/components/ExtendedTsConfig.vue
@@ -1,0 +1,34 @@
+<template>
+  <div>
+    {{ exclamationMarks }}
+    <type-script-child />
+  </div>
+</template>
+
+<script lang="ts">
+import TypeScriptChild from './TypeScriptChild.vue'
+
+import moduleRequiringEsModuleInterop from './ModuleRequiringEsModuleInterop'
+
+// The default import above relies on esModuleInterop being set to true in order to use it from
+// an import statement instead of require. This option is configured in the tsconfig.base.json,
+// so if we are no longer fully processing the tsconfig options (extended from a base config)
+// this test should fail. This was one of the only reliable ways I could get a test to fail if
+// these conditions are not being met and happen to be the use-case which was triggering errors
+// in my config setup.
+
+if (moduleRequiringEsModuleInterop()) {
+  throw new Error('Should never hit this')
+}
+
+export default {
+  computed: {
+    exclamationMarks(): string {
+      return 'string'
+    }
+  },
+  components: {
+    TypeScriptChild
+  }
+}
+</script>
--- a/packages/vue2-jest/lib/utils.js
+++ b/packages/vue2-jest/lib/utils.js
@@ -70,12 +70,18 @@ const getBabelOptions = function loadBabelOptions(filename, options = {}) {
   return loadPartialConfig(opts).options
 }
 
+const tsConfigCache = new Map()
+
 /**
  * Load TypeScript config from tsconfig.json.
  * @param {string | undefined} path tsconfig.json file path (default: root)
  * @returns {import('typescript').TranspileOptions | null} TypeScript compilerOptions or null
  */
 const getTypeScriptConfig = function getTypeScriptConfig(path) {
+  if (tsConfigCache.has(path)) {
+    return tsConfigCache.get(path)
+  }
+
   ensureRequire('typescript', ['typescript'])
   const typescript = require('typescript')
 
@@ -103,12 +109,16 @@ const getTypeScriptConfig = function getTypeScriptConfig(path) {
 
   const compilerOptions = parsedConfig ? parsedConfig.options : {}
 
-  return {
+  const transpileConfig = {
     compilerOptions: {
       ...compilerOptions,
       module: typescript.ModuleKind.CommonJS
     }
   }
+
+  tsConfigCache.set(path, transpileConfig)
+
+  return transpileConfig
 }
 
 function isValidTransformer(transformer) {
--- a/packages/vue2-jest/lib/utils.js
+++ b/packages/vue2-jest/lib/utils.js
@@ -1,6 +1,8 @@
+const ensureRequire = require('./ensure-require')
+const throwError = require('./throw-error')
 const constants = require('./constants')
 const loadPartialConfig = require('@babel/core').loadPartialConfig
-const { loadSync: loadTsConfigSync } = require('tsconfig')
+const { resolveSync: resolveTsConfigSync } = require('tsconfig')
 const chalk = require('chalk')
 const path = require('path')
 const fs = require('fs')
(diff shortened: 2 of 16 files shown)


thebanjomatic:
@lmiller1990 I wouldn't mind back-patching this to vue-jest v28.x also, but there isn't a branch to target for that. What is the process for backpatching? I support infrastructure that wraps around these tools and I'm currently having to workaround this problem in my project's v28 branch, so it would be nice to be able to remove the workaround in both places.

packages/vue2-jest/lib/utils.js
```diff
@@ -68,23 +70,55 @@ const getBabelOptions = function loadBabelOptions(filename, options = {}) {
...
-  const tsconfig = loadTsConfigSync(process.cwd(), path || '')
-  if (!tsconfig.path) {
+  if (path in tsConfigCache) {
+    return tsConfigCache[path]
+  }
+
+  ensureRequire('typescript', ['typescript'])
+  const typescript = require('typescript')
+
+  const tsconfigPath = resolveTsConfigSync(process.cwd(), path || '')
```
thebanjomatic, line 88: Note: we are still using the tsconfig package to resolve the path to the tsconfig file in the same way that `tsconfig.loadSync` was previously doing.

packages/vue2-jest/lib/utils.js
```diff
@@ -68,23 +70,55 @@ const getBabelOptions = function loadBabelOptions(filename, options = {}) {
...
+    {
+      ...typescript.sys,
+      onUnRecoverableConfigFileDiagnostic: e => {
+        const errorMessage = typescript.formatDiagnostic(e, {
+          getCurrentDirectory: () => process.cwd(),
+          getNewLine: () => `\n`,
+          getCanonicalFileName: file => file.replace(/\\/g, '/')
+        })
+        warn(errorMessage)
+      }
```
thebanjomatic, line 106: This will provide us with an error message if typescript is unable to read the tsconfig file or one of the extended files because they don't exist or don't parse etc.

packages/vue2-jest/lib/utils.js
```diff
@@ -68,23 +70,55 @@ const getBabelOptions = function loadBabelOptions(filename, options = {}) {
...
  * Load TypeScript config from tsconfig.json.
  * @param {string | undefined} path tsconfig.json file path (default: root)
  * @returns {import('typescript').TranspileOptions | null} TypeScript compilerOptions or null
  */
 const getTypeScriptConfig = function getTypeScriptConfig(path) {
-  const tsconfig = loadTsConfigSync(process.cwd(), path || '')
-  if (!tsconfig.path) {
+  if (path in tsConfigCache) {
+    return tsConfigCache[path]
+  }
```
thebanjomatic, line 83: Note: Return early from cache if we have already encountered that path.

I don't know how expensive `ts.getParsedCommandLineOfConfigFile` is, or if typescript is doing any caching internally, but I typically only call this once per-project rather than per source file when I've used it in the past. 

Since `getTypescriptConfig` is also called for each interpolated script block in the template as well as the `<script>` block of the SFC, it seemed like a good idea to just cache the parsed config result and reuse it.

packages/vue2-jest/lib/utils.js
```diff
@@ -68,23 +70,55 @@ const getBabelOptions = function loadBabelOptions(filename, options = {}) {
...
+      }
+    }
+  )
+
+  const compilerOptions = parsedConfig ? parsedConfig.options : {}
+
+  const transpileConfig = {
+    compilerOptions: {
+      ...compilerOptions,
+      module: typescript.ModuleKind.CommonJS
```
thebanjomatic, line 115: Note: As in the old code, we override the module with `commonjs`. I changed this to use the enum instead of a string literal because it failed to compile in my typescript version of this code. The expected value of `CompilerOptions['module']` is `ModuleKind`, but it does seem to accept the string literals at runtime.

packages/vue3-jest/lib/ensure-require.js
```diff
@@ -1,4 +1,4 @@
-const throwError = require('./utils').throwError
+const throwError = require('./throw-error')
```
thebanjomatic, line 1: Note: In order to use ensure-require from within utils.js, I needed to break the circular dependency. This was done by moving throw-error out of utils and into its own module so that ensure-require doesn't depend on utils.

e2e/2.x/basic/components/ExtendedTsConfig.vue
```diff
@@ -0,0 +1,34 @@
...
+import TypeScriptChild from './TypeScriptChild.vue'
+
+import moduleRequiringEsModuleInterop from './ModuleRequiringEsModuleInterop'
+
+// The default import above relies on esModuleInterop being set to true in order to use it from
+// an import statement instead of require. This option is configured in the tsconfig.base.json,
+// so if we are no longer fully processing the tsconfig options (extended from a base config)
+// this test should fail. This was one of the only reliable ways I could get a test to fail if
+// these conditions are not being met and happen to be the use-case which was triggering errors
+// in my config setup.
```
thebanjomatic, line 18: In my project, we were setting esModuleInterop in a shared base config, and `Vue.extend(...)` was throwing a TypeError when running the test because it "could not read property 'extend' from `undefined`".

thebanjomatic:
@lmiller1990 Have you had the chance to look at this PR?

lmiller1990:
@thebanjomatic I was offline for all of Oct. I'm back now, let me take a look.

e2e/2.x/basic/components/ExtendedTsConfig.vue
```diff
@@ -0,0 +1,34 @@
...
+
+if (moduleRequiringEsModuleInterop()) {
+  console.log('Should never hit this')
```
lmiller1990, line 21: In this case why not `throw Error('This should never happen')`?

packages/vue2-jest/lib/utils.js
lmiller1990: Nice

packages/vue2-jest/lib/utils.js
```diff
@@ -68,23 +70,55 @@ const getBabelOptions = function loadBabelOptions(filename, options = {}) {
   return loadPartialConfig(opts).options
 }
 
+const tsConfigCache = {}
```
lmiller1990, line 73: Hm, how about using `Map` here? Perhaps a bit cleaner and more idiomatic?

```js
const tsConfigCache = new Map()

// later

```js
let tsConfig

if (tsConfig = tsConfigCache.get(path)) {
  return tsConfig
}
```

etc.

packages/vue2-jest/lib/utils.js
lmiller1990: This is nice, I like good error messages.

lmiller1990 (commented):
This mostly looks good to me. I left a few nits - I'll give you a moment to respond, then we can merge this up and ship as 29.x.

We can look to do a 28.x back merge... I guess we will just fork from the v28 tag (I think we have one?) and back merge.

thebanjomatic:
Thanks @lmiller1990 I have updated this PR with your suggested changes. I'm no longer personally in need of the jest 28 release unless you'd just prefer to patch it because it was broken in v28.1.0

lmiller1990:
Done https://github.com/vuejs/vue-jest/releases/tag/v29.2.0

LMK if you have any issues.

PR #516, opened 2022-12-05
fix: Map lines for script block

wolfgangwalther:
I'm using vue2-jest and code coverage is broken for a while now. I noticed a difference between `processScript` in vue2-jest vs. vue3-jest:

vue2-jest:

https://github.com/vuejs/vue-jest/blob/dbe8022c59407b7474e78d2514c317685680c15e/packages/vue2-jest/lib/process.js#L35-L53

vue3-jest:

https://github.com/vuejs/vue-jest/blob/dbe8022c59407b7474e78d2514c317685680c15e/packages/vue3-jest/lib/process.js#L30-L49

The call to `mapLines` is missing for vue2-jest.

Adding the call, I can see two differences in the tests:
- The changed snapshot in the diff. Previously the `<template>` and `<style>` block didn't show up at all in the html coverage output. Now they do.
-  Looking for example at the `e2e/2.x/basic/coverage/lcov-report/basic/components/TemplateString.vue.html` file, I see the covered lines now to be in the script block as expected. Before this change they were put somewhere on the template block, which didn't make sense.

The change also made code coverage in my repo work properly again, at least I think so from the output I get.

### DIFF
--- a/e2e/2.x/basic/__snapshots__/test.js.snap
+++ b/e2e/2.x/basic/__snapshots__/test.js.snap
@@ -3,25 +3,35 @@
 exports[`generates source maps for .vue files 1`] = `
 {
   "file": "./components/Basic.vue",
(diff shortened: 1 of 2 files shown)


lmiller1990 (approved):

lmiller1990:
Seems fine, thanks!

lmiller1990:
It's out in 29.2.2. https://github.com/vuejs/vue-jest/releases/tag/v29.2.2

PR #373, opened 2021-08-10
fix: allow importing relative paths in global resources

pmrotule:
Closes #374

Fix importing relative paths inside a stylesheet being made globally available using the option `jest.globals['vue-jest'].resources`:

```json
{
  "jest": {
    "globals": {
      "vue-jest": {
        "resources": {
          "scss": [
            "./src/styles/settings/global.scss"
          ]
        }
      }
    }
  }
}
```
**src/styles/settings/global.scss**
```scss
@import './unit.scss';

$another-variable: 10px;
```

considering that _src/styles/settings/unit.scss_ exists, it shouldn't throw the following error

image

### Version

My branch is based on the `v4.0.1` tag which is the version I'm using. It doesn't seem like there is a `v4` branch I could base my PR on.

PR #548, opened 2023-08-10
fix: allow importing relative paths in global resources

pmrotule:
It seems like the fix I've made for relative paths in global resources in vue2-jest v26 (https://github.com/vuejs/vue-jest/pull/373) has not been ported to the other versions (not in v27, v28, v29). We recently upgraded to v29 and faced the same issue so this PR is making the same changes, but for v29 in both Vue 2 and Vue 3 (I tested it and it works as expected).

More details found in my previous PR: https://github.com/vuejs/vue-jest/pull/373

### DIFF
--- a/e2e/2.x/style/colors.less
+++ b/e2e/2.x/style/colors.less
@@ -1,0 +1,1 @@
+@primary-color: "red";
--- a/e2e/2.x/style/colors.scss
+++ b/e2e/2.x/style/colors.scss
@@ -1,0 +1,1 @@
+$primary-color: #333;
--- a/e2e/2.x/style/variables.less
+++ b/e2e/2.x/style/variables.less
@@ -1,1 +1,3 @@
-@primary-color: "red";
+@import "./colors.less";
+
+@font-size: 16px;
--- a/e2e/2.x/style/variables.scss
+++ b/e2e/2.x/style/variables.scss
@@ -1,1 +1,3 @@
-$primary-color: #333;
+@import './colors.scss';
+
+$font-size: 16px;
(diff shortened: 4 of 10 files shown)


pmrotule:
@lmiller1990 Would you have time to have a look?

lmiller1990:
Will release this one now.

lmiller1990:
Done: https://github.com/vuejs/vue-jest/releases/tag/v29.2.6

remcobeunk:
@pmrotule it seems this fix is breaking our unit test run with the error you originally fixed in this pr: https://github.com/vuejs/vue-jest/pull/373.

On Mac it works but on Windows machines it does not.

Error: Can't find stylesheet to import.
  ╷
1 │ @import "XXXX path = correct and file exists";
  │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ╵
  
After reverting your change locally our vue2 tests start working again:
image

Any ideas?

pmrotule:
@remcobeunk I would need more information to debug.
- Which CSS preprocessor do you use?
- How does your global ressources config look like?

Can you provide a reproduction repo or stackblitz?

remcobeunk:
Our project uses SASS CSS pre-processor 
Global resources config? No idea which config you want, jest specifically or all configs?
But because project is end of live we are not going to invest more time into this problem. 

We set our project fixed to version of 29.2.5 that works for all of our developers. 
Sorry I cannot help you more with this.

